PR #36, opened 2017-02-06
type coneversion to float when checking adcgain 

amit4111989:
for condition 
if adcgain =0 ; then
    adcgain=200

adcgain is returned as string and therefore type conversion to float is needed when checking for a 0 value


cx1111:
Thanks, I should have fixed this in the new refactor.

Issue #40, opened 2017-04-11
Downloads fail on Windows


kbrose:
Main problem is prevalence of using `os.path.join()` to create URLs, on Windows this will use backslashes to join the arguments instead of forward slash, ending up with errors like

```
>>> import wfdb; import os; wfdb.dldatabasefiles('ltstdb', os.getcwd(), ['s20011.hea', 's20011.dat'], overwrite=True)
Downloading files...
multiprocessing.pool.RemoteTraceback:
... < SNIP > ...
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://physionet.org/physiobank/database/ltstdb%5Cs20011.hea
```

(of course `%5C` is a backslash).

Probably want to use `posixpath` instead?

gloryren:
> Main problem is prevalence of using `os.path.join()` to create URLs, on Windows this will use backslashes to join the arguments instead of forward slash, ending up with errors like
> 
> ```
> >>> import wfdb; import os; wfdb.dldatabasefiles('ltstdb', os.getcwd(), ['s20011.hea', 's20011.dat'], overwrite=True)
> Downloading files...
> multiprocessing.pool.RemoteTraceback:
> ... < SNIP > ...
> requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://physionet.org/physiobank/database/ltstdb%5Cs20011.hea
> ```
> 
> (of course `%5C` is a backslash).
> 
> Probably want to use `posixpath` instead?

have solved this?
i have this problem now, and i check WFDB have use posixpath now ,but why still this happened?

PR #42, opened 2017-04-11
Windows compatibility

kbrose:
Adding windows compatibility to download commands.

See #40.

### DIFF
--- a/wfdb/downloads.py
+++ b/wfdb/downloads.py
@@ -4,31 +4,31 @@ import os
 import sys
 import requests
 from IPython.display import display
-        
+
 # Read a header file from physiobank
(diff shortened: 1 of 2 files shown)


kbrose:
This gets me down from 13 errors to 2 errors when running `nosetests` on windows. The 2 remaining errors are truly mysterious to me (problems with `_headers.inttypes`). If this gets merged then I'll make an Issue for those and try and get them fixed.

kbrose:
Still not quite sure why the part fixed in ed8fcb6 wasn't working on my machine, but I believe the change should be a little more robust anyway. With that commit I have no errors when running `nosetests` locally.

cx1111:
Thanks for your help.

PR #59, opened 2017-06-07
Add numeric representation of annotations

roszcz:
Access to this information as int can make automated analysis simpler.


cx1111:
I've added options to return the annotation labels in 3 possible formats/variables: `label_store`, `symbol`, `description`. You can use the `return_label_elements` input variable to specify which ones you want returned.

PR #61, opened 2017-06-20
Use isinstance() for argument type checking

jcbsv:


cx1111:
Currently rewriting the annotations api. Thanks for your pull request.

Could you tell me why isinstance is preferred to type()? I know:
- isinstance works for child objects.
- isinstance looks more elegant and makes the code more understandable.

Is there anything else?

jcbsv:
The normal, Pythonic, preferred solution is almost invariably "duck typing": try using the argument as if it was of a certain desired type. But if you insist on explicit type checking, isinstance is preferable over type because it seamlessly supports inheritance/sub-classing. See, e.g., this SO answer for more details: https://stackoverflow.com/questions/1549801/differences-between-isinstance-and-type-in-python

cx1111:
I've made the changes in my most recent update. Thanks.

Issue #151, opened 2018-12-12
Wrong definition of specificity 


ilupin:
https://github.com/MIT-LCP/wfdb-python/blob/cc8c9e9e44f10af961b7a9d8ae03708b31ac8a8c/wfdb/processing/evaluate.py#L246

Hi, the value above should be sensitivity by [definition](https://en.wikipedia.org/wiki/Sensitivity_and_specificity).

PR #158, opened 2019-02-01
Dev signalpy

Gabriel-Azevedo-Ferreira:

PR #159, opened 2019-02-01
Dev evaluate

Gabriel-Azevedo-Ferreira:
Solving the problem appointed in this thread:
https://github.com/MIT-LCP/wfdb-python/issues/151


Lucas-Mc:
This one has the same changes as PR #158 .. what is going on here? Which one is correct?

Issue #163, opened 2019-03-02
easify xlabel adaption for plotting


matu3ba:
Why is there no option to set the x-axis via `xlabel` for `plot_items` ?
Trying to adapt the figure via
ylabels=['mV']
fig = plot_items(signal=signal, ylabel=ylabels, return_fig=True)
fig.xlabel = 'test'
fig.savefig('plot' + str(plotcount) + '.eps', format='eps')
works without the change of xlabel.
https://stackoverflow.com/questions/12444716/how-do-i-set-the-figure-title-and-axes-labels-font-size-in-matplotlib

Am I missing anything or is there any technical reason why to not give the option to overwrite the axis?
You could also return the axis in addition to the figure to make adaptions possible.
The saving of .eps format upon dialog may be broken as well, but that is not big deal.

Dubrzr:
Hi matu3ba, 
Don't hesitate to make pull requests if you think something is wrong/missing.
Julien

PR #164, opened 2019-03-04
Plotadaption

matu3ba:
This change makes xlabels of matplotlib accessible for user change and gives possibility to overwrite the xlabel by additional parameter.
As such, it fixes/expands #163 
Original file did not contain test for plot_items as well, but I did test with a data set of my project, so I can assure it works.

### DIFF
--- a/wfdb/plot/plot.py
+++ b/wfdb/plot/plot.py
@@ -10,8 +10,8 @@ from ..io.annotation import Annotation
 
 def plot_items(signal=None, ann_samp=None, ann_sym=None, fs=None,
                time_units='samples', sig_name=None, sig_units=None,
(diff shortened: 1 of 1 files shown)


Lucas-Mc:
@matu3ba This looks like it's working well and ready to merge, but I am confused about the `return_fig_axes` section. Once we resolve this, I think it's ready though since it worked locally. 

Some further considerations would be to add error handling where the user might accidentally input either too many or too little x-labels. Too much wouldn't result in a error I believe, but would produce unexpected behavior from what they wanted. Too little definitely would produce an error since the following line would return an out of bounds index. 
https://github.com/MIT-LCP/wfdb-python/blob/49f0b1db0149e95889df8a5685f40b39053eec46/wfdb/plot/plot.py#L336

PR #200, opened 2020-04-29
Replaces Physiobank with Physionet #199

### LINKED ISSUE #199: Replace Physiobank with Physionet
Physiobank is not a permanent resource for waveform data and this package should not rely on it to retrieve its data. Physionet is the updated version of Pysiobank (at some point it will completely replace Physiobank and Physiobank will no longer exist) and it includes all of Physiobank as well as some content unique to itself. For this reason, all the references to Physiobank in this package should be updated to point to the latest storage space for the content.

Lucas-Mc:
Replaces (nearly) all instances of a call to Physiobank with the equivalent call to Physionet. The only instance that hasn't been replaced right now is the list of all available projects, since this functionality has not been set up for Physionet yet. In addition, the Jupiter Notebook demo has been edited and tested to run with the new content. Test files have been updated. All references to Physiobank in the docstrings and variable names has also been changed. Further commits might completely eliminate any data files which are solely in Physiobank and not in Physionet as these would become deprecated upon Physiobank's removal. Fixes #199.


Lucas-Mc:
Documentation (Read the Docs) will have to be updated as well. References to the original WFDB documentation is quickly becoming deprecated and outdated and needs to be updated to prevent misleading examples or examples which need tweaking to work.

PR #218, opened 2020-05-22
Produces record file from EDF format #87

### LINKED ISSUE #87: How to read a *.edf file with rdsamp?
Hello!
    Well,I recently have a problem when using the wfdb-python.While reading a *.edf file the program will encounter errors and it will remind me that there is no *.edf.hea file.Em...I don't konw how to solve this problem.
    Look forward to your reply,sincerely.

Lucas-Mc:
Introduces the conversion of EDF files to MIT format. Utilizes [PyEDFlib](https://pyedflib.readthedocs.io/en/latest/) to read in the EDF files then conversion was done after that to MIT format. I thought about using [MNE](https://mne.tools/stable/generated/mne.io.read_raw_edf.html), but this is a fairly large package and their EDF conversation dictionary was more difficult to understand in my opinion. 

Further work plans to generate the `.dat` files. I've tested it already and can generate the `.hea` files but it looks like the `.dat` files will need a little more work that I should just do separately. 

Also provides some testing framework to check differences between records imported after running the original `edf2mit` command and those generated from reading the EDF file using this code here in Python. Fixes #87.


Lucas-Mc:
Just adding a note that I would like to add some documentation regarding the format of the EDF signals as they are converted to MIT format for future reference while developing. The arrays are non-intuitive though make sense once the objective of the conversion process is understood.

Lucas-Mc:
Sometimes, `PyEDFlib` outputs this error: `OSError: the file is not EDF(+) or BDF(+) compliant`. This issue is discussed more [here](https://github.com/holgern/pyedflib/issues/14). It appears like `PyEDFlib` is pretty strict in their formatting requirements that MNE isn't. I think, for this reason, I will switch MNE instead to allow for increased compatibility with different formats (EDF, EDF+, BDF, etc.).

Lucas-Mc:
Tested this with the most recent pull request #226 to determine time saving with reading EDF files and found a half time speed increase with limited signal accuracy loss.
Screen Shot 2020-06-04 at 9 48 31 AM
Screen Shot 2020-06-04 at 9 48 53 AM

alistairewj:
What do you mean "accuracy loss"? Is this just floating point imprecision or actual information loss? From what I remember EDF is 16 bit so `float16` should be able to retain all the information.

Issue #225, opened 2020-06-01
Add option for user to adjust numpy datatype


Lucas-Mc:
Add parameter in `rdrecord`/`rdsamp` for user to select their desired numpy datatype since the default now is `float64` and this appears to cause `MemoryError` with larger datasets. See #224.

It should be noted that `rdrecord` currently has a `return_res` parameter though this only converts the `dtype` after the data has already been imported and manipulated in the `float64` format.

PR #226, opened 2020-06-02
Adds datatype parameter in rdrecord/rdsamp #224 #225

### LINKED ISSUE #224: MemoryError
While I'm trying to read waveform data using rdsamp / rdrecord, I encounter this 'MemoryError' issue many a times. I'm using Python 64 bit and am clueless on how to solve it. Any leads?

Lucas-Mc:
Adds a datatype parameter in the `rdrecord` and `rdsamp` functions to allow the user to increase computation speed at the expense of accuracy and significant figures of signal data. Fixes #224 and #225.

### DIFF
--- a/wfdb/io/_signal.py
+++ b/wfdb/io/_signal.py
@@ -856,9 +856,10 @@ class SignalMixin(object):
 
 #------------------- Reading Signals -------------------#
 
+
 def _rd_segment(file_name, dir_name, pn_dir, fmt, n_sig, sig_len, byte_offset,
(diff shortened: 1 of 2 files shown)


Lucas-Mc:
Here are some raw data results to support switching the output `numpy` datatypes (in this case `float64` to `float16`. Further savings can probably be made by switching to `int8` if possible.

1. The computation time is cut in half:

Screen Shot 2020-06-02 at 11 53 12 AM

2. The cost in signal accuracy is extremely minimal and can be ignored:

Screen Shot 2020-06-02 at 11 53 32 AM

PR #241, opened 2020-06-20
Produces MAT file from WFDB format

Lucas-Mc:
Introduces the conversion of WFDB files to MAT format. All data was written to the `.mat` file using `struct.pack()` so no outside packages were introduced.

This implementation conserves the `-f` and `-t` options for setting a desired start and stop time respectively, though the units here are samples instead of seconds. The `-h` option can be used using `help(wfdb.wfdb2mat)`. The `-H` option can not be implemented here at the moment since all high-resolution content is read first through the `rdrecord` function which must be changed to allow this. The `-l` option can be used with the `-f` and `-t` options in combination. The `-s` option is implemented by allowing the user to select their desired channels based on the integer value of the signal number. The `-S` option was not implemented here due to its limited potential use. 

Furthermore, I added functionality to write the `.hea` file as well to accompany the generation of the MAT file so that it can be read again with `rdrecord` and complete the cycle!

### DIFF
--- a/wfdb/__init__.py
+++ b/wfdb/__init__.py
@@ -1,5 +1,5 @@
 from wfdb.io.record import (Record, MultiRecord, rdheader, rdrecord, rdsamp,
-                            wrsamp, dl_database, edf2mit, wav2mit, sampfreq, signame)
+                            wrsamp, dl_database, edf2mit, wav2mit, wfdb2mat, sampfreq, signame)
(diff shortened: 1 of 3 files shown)


Lucas-Mc:
Something else cool to note is that a MAT file can now be created from an EDF or WAV file since `rdrecord` now allows reading of those formats into WFDB format!

PR #243, opened 2020-07-06
Add linting and formatting standards

cx1111:
- Add config files for `pylint` and `black`
- Apply `black`
- Clean up/recorder some imports


wfdb/io/record.py
```diff
@@ -1921,42 +2187,68 @@ def wfdb2mat(record_name, pn_dir=None, sampfrom=0, sampto=None, channels=None):
...
     for i in range(record.n_sig):
-        print('{:<7}{:<20}{:<17}{:<10}{:<10}'.format(i,
-                                                     record.sig_name[i],
-                                                     record.adc_gain[i],
-                                                     record.baseline[i]-record.adc_zero[i]+offset,
-                                                     record.units[i]))
+        print(
+            "{:<7}{:<20}{:<17}{:<10}{:<10}".format(
+                i,
+                record.sig_name[i],
```
Lucas-Mc, line 2246: Is this really easier to read than the previous version?

tests/test_processing.py
```diff
@@ -60,63 +64,139 @@ def test_find_peaks_empty(self):
...
+            1181,
+            1469,
+            1770,
+            2055,
+            2339,
+            2634,
+            2939,
+            3255,
+            3551,
+            3831,
```
Lucas-Mc, line 88: Is this the PEP8 way to do this? What's wrong with the other one if it's under 80 characters?

bemoody:
I don't particularly want to spend any time on this, but you should go ahead with reformatting if you think it'll be beneficial.

I have half a dozen or so unfinished branches (and I'm not going to stop all work on them to wait for this to be done), but that doesn't mean such changes can't be rebased.

PR #261, opened 2020-11-03
Prune requirements

cbrnr:
I just did `pip install wfdb` and noticed that there are some dependencies that can probably be removed. The package `sklearn` is not a real package (you want `scikit-learn`) so it should be removed. I've also moved `nose` to `extras_require` because it is not required for normal usage (I've kept it in `requirements.txt` though). Finally, I think `six` is never used in the project so I've removed it. Not sure if there are more unused dependencies, but these are probably the most important fixes. Let me know what you think.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,5 @@ pytz==2020.1
 requests==2.23.0
 scikit-learn==0.22.2.post1
 scipy==1.4.1
-six==1.15.0
 threadpoolctl==2.1.0
 urllib3==1.25.9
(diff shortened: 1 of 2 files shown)


Lucas-Mc:
Hey @cbrnr! Thanks for the input! It appears like `six` is used by `cycler` upon install so we should keep it pinned down. Good catch on the rest! Yeah `sklearn` has been on the long list of things to get around to haha and I agree about `nose` since it's only used for testing e.g. `pip install .[test]` then `nosetests`

cbrnr:
Cool, thanks for the quick response! Regarding `six`, I still think that it's not necessary to pin dependencies of dependencies - `cycler` will take care of that.

PR #264, opened 2020-11-13
Add option to not use multiprocess

lbugnon:
Adding a parameter to make optional to use mutliprocess. I found errors when using miniconda in windows with this implementation and files cannot be downloaded. Maybe there are other workaorunds, but this one works well and it is optional (by default, multiprocess is used)

### DIFF
--- a/wfdb/io/download.py
+++ b/wfdb/io/download.py
@@ -278,7 +278,9 @@ def get_record_list(db_dir, records='all'):
 
     """
     # Full url PhysioNet database
-    if os.sep not in db_dir:
+    #if os.sep not in db_dir:
+    print(db_dir)
+    if "/" not in db_dir:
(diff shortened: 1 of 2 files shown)


Lucas-Mc:
Thanks! Can you rebase this on the current `master` branch and remove the commented-out old code?

lbugnon:
@Lucas-Mc there it is, thanks!

lbugnon:
@Lucas-Mc there it is, thanks!

bemoody:
Sorry I hadn't looked at this before.  Just a couple of quick comments:

- Have you tried replacing `multiprocessing.Pool` with `multiprocessing.dummy.Pool`?  That might solve your problem.

- Please don't add `print` statements unconditionally (in fact, it's best to avoid them entirely.)  You might consider adding a `verbose` argument if you want to display status info on stdout, but a better choice might be to use the `logging` module.

Issue #269, opened 2020-12-18
AttributeError: 'NoneType' object has no attribute 'shape' for wfdb.plot_wfdb call with an annotation object but no record object


briangow:
This call will result in the error: 

```
annotation = wfdb.rdann('100', 'atr', sampfrom=100000, sampto=110000,
pn_dir='mitdb')
annotation.fs = 360
wfdb.plot_wfdb(annotation=annotation, time_units='minutes')
```

as shown here:

```
wfdb/plot/plot.py in 
plot_items(signal, ann_samp, ann_sym, fs, time_units, sig_name, 
sig_units, xlabel, ylabel, title, sig_style, ann_style, ecg_grids, 
figsize, return_fig,
return_fig_axes)
     129
     130     if ylabel:
 --> 131         if len(ylabel) != signal.shape[1]:
     132             raise Exception('The length of the ylabel must be the
 same as the '
     133                             'signal: {}
 values'.format(signal.shape[1]))
AttributeError: 'NoneType' object has no attribute 'shape'
```

The wfdb.plot_wfdb function doesn't require a record object but if it isn't passed (and therefore there isn't a signal object), this error will occur if there is a ylabel along with an annotation object.

PR #272, opened 2020-12-22
Fix for MIT-LCP/wfdb-python issue #269. The number of subplots (n_sub…

briangow:
Fix for MIT-LCP/wfdb-python issue #269. The number of subplots (n_subplots) should match the number of ylabels regardless of whether records or annotations are being passed in.


Lucas-Mc (approved):
Looks good!

PR #275, opened 2021-01-05
Relaxes requirements for pytz

Lucas-Mc:
This change relaxes the requirements for the `pytz` package from `>=2018.3` to `>= 2017.2` since it is only required by `pandas` which only requires `pytz>=2017.2`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setup(
         'pandas>=0.17.0',
         'pyparsing>=2.0.4',
         'python-dateutil>=2.4.2',
-        'pytz>=2018.3',
+        'pytz>=2017.2',
         'requests>=2.8.1',
         'scikit-learn>=0.18',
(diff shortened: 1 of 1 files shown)


tompollard:
Could you explain why it is necessary to relax the requirement? What is the pytz package and which version are we running on the server?

Lucas-Mc:
From the most recent PhysioNet build:
```
The conflict is caused by:
    The user requested pytz==2017.2
    django 2.2.13 depends on pytz
    google-api-core 1.13.0 depends on pytz
    pandas 1.1.0 depends on pytz>=2017.2
    wfdb 3.1.2 depends on pytz>=2018.3
```
That original requirement was too strict since when I removed it from `requirements.txt` for WFDB-Python only `>=2017.2` was required. The version should be `2017.2` based on the `requirements.txt` for PhsyioNet.

Issue #260, opened 2020-10-23
wrann not working with only label_store set in Annotation


Lullatsch:
Hi I think the `wrann` method inside `wfdb.io.Annotation` might not be working as intended right now. 

When I try to write a new annotation with only the label_store field set, I get a Nonetype error inside `wfdb.io.Annotation.calc_core_bytes` (line 930) where it says `data_bytes.append(field2bytes('samptype', [sampdiff[i], self.symbol[i]]))`, because self.symbol is None (and thus unsubscriptable). 

From the documentation it should however also be possible to only use label_store as a field in my Annotation.

Tomwhiting1987:
I'm also getting exactly the same error when I put in a label store array instead of a symbol array. I thought it was something silly I was doing! Is it possible to fix this? It does suggest in the documentation that this should work.

PR #279, opened 2021-02-23
Fixes NoneType error in wfdb.io.wrann when using label_store instead of symbol #260

briangow:
`symbol` needs to be defined in `calc_core_bytes` so the annotations can be converted into bytes for writing.  I added this to define `symbol` when `label_store` is being used:

```
 # Calculate the symbol field if necessary
        if 'symbol' not in present_label_fields:
            self.convert_label_attribute(source_field=present_label_fields[0],
                                         target_field='symbol')
```


Lucas-Mc (commented):
Syntax looks good and consistent and it works!

PR #280, opened 2021-02-24
Adds CSV to WFDB converter

Lucas-Mc:
This change adds a CSV to WFDB converter. This function only explicitly requires the sample rate (`fs`) and units (`units`) to be defined. The rest of the information will be interpreted based on the signal provided and common default values unless explicitly stated by the user attempting the conversion. The user may either choose to only return the record (`record_only=True`), write the file (nothing must be explicitly stated as this is the default behavior), turn on verbosity to display the computed values to adjust if needed (`verbose=True`), set a custom delimiter (`delimiter='\t'`), and specify if a header is being used (`header=False`).

In addition to the CSV being required to follow [RFC guidelines](https://tools.ietf.org/html/rfc4180#section-1), it must also adhere to the following format as well:
```
sig_1_name,sig_2_name,...
sig_1_val_1,sig_2_val_1,...
sig_1_val_2,sig_2_val_2,...
...,...,...
```

### DIFF
--- a/docs/io.rst
+++ b/docs/io.rst
@@ -9,7 +9,7 @@ WFDB Records
 ---------------
 
 .. automodule:: wfdb.io
-    :members: rdrecord, rdheader, rdsamp, wrsamp, edf2mit, mit2edf, wav2mit, mit2wav
+    :members: rdrecord, rdheader, rdsamp, wrsamp, edf2mit, mit2edf, wav2mit, mit2wav, csv2mit
 
 .. autoclass:: wfdb.io.Record
     :members: wrsamp, adc, dac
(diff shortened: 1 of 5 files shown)


briangow:
This seems very helpful!  I have checked this on a CSV file which has a very high channel count.  It works great!  The code to automatically duplicate a parameter such as `units` across all channels if it is passed as a `str` (or `int` in other cases) instead of a list saves another step for users when all of the channels have the same units.  I do think we should update the description for the parameters that this new behavior applies to and clearly describe this option and how it works.  

Also, most wfdb functions, including this one, don't allow for a save path and instead save to the working directory.  In the future, it may be nice to allow the user to save to a specific path for functions which perform saves.

Lucas-Mc:
Thanks for the comments! I think we could add the saving to a new path functionality in the separate pull request which would add a new parameter to apply to all functions so that they can specify which path they would like to save the files in

briangow:
That sounds like a good plan.

PR #287, opened 2021-03-31
Adds wfdbdesc function

Lucas-Mc:
This change adds the `wfdbdesc` function from the [original WFDB package](https://www.physionet.org/physiotools/wag/wfdbde-1.htm). The only difference here is that I chose to limit the description of multi-segment headers to just the header itself instead of diving into each of the records it calls (I thought this might be a bit more consistent behavior). This can be added as a feature later though.


Lucas-Mc:
The tests are failing since I used f-strings which are Python 3.6+ but Travis is running on 3.5

Issue #73, opened 2017-09-26
Maybe wfdb should not read .{dat,hea,...}, instead it should be possible to directly give wfdb file contents.


Dubrzr:
The fact that wfdb read files for you and that you don't have to open the file is nice, but when your files are not on the local filesystem it is much more complicated to open those files with wfdb.
For example:
Let's say all my files are on Hadoop HDFS (which is actually the case), the only way to access them in python is using a library such as hdfs3 (https://github.com/dask/hdfs3):

```
with hdfs.open('/user/data/100.dat') as f:
    content = f.read()
# I am now forced by wfdb to write the contents to a temporary file in the local filesystem to then open it with wfdb...
```
I know it would be very annoying to modify the current implementation of wfdb but I have the feeling it would be better to let the user choose its own way to give .{dat, hea, ...} contents to wfdb.
The way I see this:
1. You have the wfdb-python library that only takes contents as input
2. You have helper functions (downloads from physionet and read the contents for you)

Do you see a better solution?

lkjell:
I might add to this that there is no way to spesify where .hea files to write to. Currently I can change the directory where .dat writes to by modifying the filename attribute of Record.

josemarfdc:
For HDFS, one can always use the [NFS gateway ](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/HdfsNfsGateway.html)and mount the required HDFS path as a NFS folder.
I'm sure there should be something similar for Object data store (AWS S3, GCS Bucket).

IMO a better solution would be the ability to write to an [IO ByteStream](https://docs.python.org/3/library/io.html) instead. This would give users a memory object that could be managed however needed (HDFS, S3, GCP objects, etc).

If the maintainers agree, I could try and pitch in some dev time.

PR #298, opened 2021-04-29
Adds Amazon AWS S3 support #268

### LINKED ISSUE #268: read records stored in AWS S3
Can we add a function to read records stored in AWS S3 like Pandas?

Lucas-Mc:
This change adds support for data stored in Amazon AWS S3 (buckets). For the moment, anywhere there is a `pn_dir` parameter in a function, the user can provide the S3 URI form which to find the desired input file. This should always begin with `'s3'` in order to work correctly. An example input would be: `'s3://my-aws-bucket/'`.

For example, I can read record `100` (both `.dat` and `.hea`) stored on `my-aws-bucket` like this:
```
>>> import wfdb
>>> rec = wfdb.rdrecord('100', pn_dir='s3://my-aws-bucket')
```

As mentioned in #73, I think it would be nice to transition away from WFDB's dependence on files and take an approach similar to this statement:
```
1. You have the wfdb-python library that only takes contents as input
2. You have helper functions (downloads from physionet and read the contents for you)
```
Where each helper function would be for a separate application (PhysioNet, AWS S3, Hadoop HDFS, etc.). This would probably be a significant change (I think `pn_dir` should be changed to `remote_dir` at least) so something to look forward to in `v4.0`??

Fixes #268.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ scikit-learn==0.22.2.post1
 scipy==1.4.1
 threadpoolctl==2.1.0
 urllib3==1.25.9
+s3fs==2021.4.0
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,8 @@ setup(
         'scikit-learn>=0.18',
         'scipy>=0.17.0',
         'threadpoolctl>=1.0.0',
-        'urllib3>=1.22'
+        'urllib3>=1.22',
+        's3fs>=2021.4.0'
     ],
 
     # List additional groups of dependencies here (e.g. development
--- a/wfdb/io/_signal.py
+++ b/wfdb/io/_signal.py
@@ -2091,7 +2091,7 @@ def _infer_sig_len(file_name, fmt, n_sig, dir_name, pn_dir=None):
         file_size = os.path.getsize(os.path.join(dir_name, file_name))
     else:
         file_size = download._remote_file_size(file_name=file_name,
-                                               pn_dir=pn_dir)
+                                               remote_dir=pn_dir)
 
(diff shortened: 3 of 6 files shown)


bemoody:
Nice to know this is possible, though I haven't tested it.

I do see a few problems with the implementation here that would need to be addressed:

 - Numerous functions check whether `remote_dir.startswith('s3')`.  This conflicts with the existing namespace (for example, `rdrecord("foo", pn_dir="s345db/1.0")` ought to read the record `http://physionet.org/files/s345db/1.0/foo`.)  If reusing the `pn_dir` API is desirable, it needs to be unambiguous (for example, `remote_dir.startswith('s3://')` instead of `remote_dir.startswith('s3')`.)

 - `_stream_dat` ignores `start_byte` and `byte_count`, which means it will give wrong results when reading a subset of a signal file, or when reading a signal file with a prolog.

 - Rather than creating abstractions to simplify the logic, this adds complexity to both the high-level `record` functions and the low-level `download` functions.

For a lot of reasons, including reducing the existing complexity, I'm trying to replace the download functions with an abstract file I/O layer (see pull #320.)

I know nothing about s3fs, but just looking at how you're using it here, I'm guessing that it already provides a BufferedIOBase interface or something similar.  Which means it ought to be simpler, and much more maintainable, to do this by building on top of pull #320 instead.

I'd welcome thoughts from anyone about what the ideal API should look like.  `pn_dir` is pretty clunky in my mind.

tompollard:
@briangow can this issue be closed? I think the work is superceded by https://github.com/MIT-LCP/wfdb-python/pull/523.

briangow:
@tompollard , yes, closing this makes sense to me.

PR #299, opened 2021-04-29
Adds extra header information

Lucas-Mc:
Every time I look at a header file I have to remind myself of what exactly the format is and what the separators (`:`, `+`, etc.) mean. Therefore, a while back, a put together what a header file would look like given that every field is filled. I decided that it would be useful for others to have this information also so I'm adding it in this pull request.


briangow:
Thanks @Lucas-Mc , this is very helpful!

PR #310, opened 2021-06-21
This updates the Matplotlib version to avoid segmentation faults on Macs running Big Sur.

### LINKED ISSUE #309: Segmentation fault on Mac with older version of Matplotlib
Matploblib version 3.2.1 gives a segmentation fault error when plotting from Matplotlib.pyplot on a Mac running Big Sur: https://stackoverflow.com/questions/64842404/matplotlib-segmentation-fault-mac-os-11big-sur .  The current version in requirements.txt is 3.2.1.

briangow:
Fixes #309

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ cycler==0.10.0
 idna==2.9
 joblib==0.14.1
 kiwisolver==1.2.0
-matplotlib==3.2.1
+matplotlib==3.3.4
 nose==1.3.7
 numpy==1.18.5
 pandas==1.0.3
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks @briangow. This change updates the version in the requirements (used for installing the development environment) but to fix the bug for people pip installing the package we should update https://github.com/MIT-LCP/wfdb-python/blob/master/setup.py

briangow:
Thanks @tompollard!  I've updated the setup.py file.

tompollard:
Looks good, thanks!

PR #317, opened 2021-08-04
Fix reading 24-bit signals (format 24)

bemoody:
WFDB format 24 consists of 24-bit signed integers (each sample is stored as three bytes on disk.)  When reading the signal data into memory, it therefore must be converted into some format that the CPU can work with, such as 16-bit or 32-bit integers.

Previously, this was done by silently discarding the least significant byte and storing the samples as 16-bit integers.  This loses precision, which is rather a bad thing if you actually have a reason for using 24-bit signal files.  Additionally, since `rdrecord` didn't make any corrections to the gain/baseline, the resulting physical values were completely wrong.

Furthermore, the conversion was done in a messy and very buggy way in `_rd_dat_file` (failing to handle `sampfrom`/`sampto`, failing to handle prologs, and failing to handle an odd number of samples, among other issues.)  All of these problems can be fixed by handling format 24 in the same way as other "unaligned" formats.

Note that *writing* format 24, in `wrsamp`, seems to work correctly.

Example:
```
$ snip -r 100s -n 100s24a -O 24 -t s10
$ rdsamp -p -r 100s24a
          0.000  -0.145  -0.065
          0.003  -0.145  -0.065
          0.006  -0.145  -0.065
          0.008  -0.145  -0.065
          0.011  -0.145  -0.065
          0.014  -0.145  -0.065
          0.017  -0.145  -0.065
          0.019  -0.145  -0.065
          0.022  -0.120  -0.080
          0.025  -0.135  -0.080
```
Using the current `master` branch:
```
$ python3 -q
>>> import wfdb
>>> wfdb.rdrecord('100s24a').p_signal
array([[-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105],
       [-5.105, -5.105]])
```
Expected result:
```
$ python3 -q
>>> import wfdb
>>> wfdb.rdrecord('100s24a').p_signal
array([[-0.145, -0.065],
       [-0.145, -0.065],
       [-0.145, -0.065],
       [-0.145, -0.065],
       [-0.145, -0.065],
       [-0.145, -0.065],
       [-0.145, -0.065],
       [-0.145, -0.065],
       [-0.12 , -0.08 ],
       [-0.135, -0.08 ]])
```

### DIFF
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -140,7 +140,7 @@ class TestRecord(unittest.TestCase):
         # Test file writing
         record_2 = wfdb.rdrecord('sample-data/n8_evoked_raw_95_F1_R9', physical=False)
         record_2.wrsamp()
-        record_write = wfdb.rdrecord('sample-data/n8_evoked_raw_95_F1_R9', physical=False)
+        record_write = wfdb.rdrecord('n8_evoked_raw_95_F1_R9', physical=False)
 
         assert np.array_equal(sig, sig_target)
         assert record.__eq__(record_pn)
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -125,11 +125,13 @@ class TestRecord(unittest.TestCase):
         Format 24, entire signal, digital.
 
         Target file created with:
-            rdsamp -r sample-data/n8_evoked_raw_95_F1_R9 | cut -f 2- > record-1e
+            rdsamp -r sample-data/n8_evoked_raw_95_F1_R9 | cut -f 2- |
+            gzip -9 -n > record-1e.gz
         """
         record = wfdb.rdrecord('sample-data/n8_evoked_raw_95_F1_R9', physical=False)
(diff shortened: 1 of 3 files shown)


bemoody:
Okay, so, yeah, test_1e fails, because test_1e is *wrong*.  `tests/target-output/record-1e` is not in any way, shape, or form the output of `rdsamp -r sample-data/n8_evoked_raw_95_F1_R9 | cut -f 2-`.

More than that, the wrsamp part of test_1e isn't actually testing what it ought to be testing - if it *were*, it would have failed.

Sorry for the noise - I'll fix this up...

tompollard:
Looks good to me, thanks!

PR #320, opened 2021-08-13
HTTP file I/O backend

bemoody:
When reading remote files via HTTP(S), currently a variety of ad-hoc methods are used: e.g., there's one function for retrieving and parsing a header file, another function for retrieving a segment of a signal file, and another for retrieving an annotation file.

This isn't conducive to reading complex data formats, because every format needs to be implemented twice (once using the standard file API, for local files, and once using the requests API, for remote files.)  This leads to bugs and inconsistencies.

Furthermore, it is desirable to reuse sockets, set a meaningful User-Agent, and retain cookies.  This requires handling all of the package's HTTP requests in a central place, rather than each function calling `requests.get` or `requests.head`.

This pull request introduces a function `openurl` which opens a remote URL as a file object.  This object implements the standard Python file API, so you can read() and seek() and do everything else that you'd do with a normal Python file.  This is analogous to the `struct netfile` used by libwfdb.

Here, I've simply modified the `wfdb.io.download` functions to use this backend instead of calling `requests` directly.  In future pull requests I'll try to remove as much as I can of the extraneous duplicated logic.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1,0 +1,322 @@
+import gzip
+import http.server
+import threading
+import unittest
+
+import wfdb.io._url
+
+
+class TestNetFiles(unittest.TestCase):
+    """
+    Test accessing remote files.
+    """
+    def test_requests(self):
+        """
+        Test reading a remote file using various APIs.
+
+        This tests that we can create a file object using
+        wfdb.io._url.openurl(), and tests that the object implements
+        the standard Python API functions for a file of the
+        appropriate type.
+
+        Parameters
+        ----------
+        N/A
+
+        Returns
+        -------
+        N/A
+
+        """
+
+        text_data = """
+        BERNARDO: Who's there?
+        FRANCISCO: Nay, answer me: stand, and unfold yourself.
+        BERNARDO: Long live the king!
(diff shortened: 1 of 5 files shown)


bemoody:
Nice work on the test suite catching all my silly mistakes!

tompollard:
@bemoody please could you rebase this on the main branch?

tompollard:
thanks @bemoody, this sounds like a big improvement in the way remote files are handled. please could you help to answer a couple of questions?

- what is the purpose of the duplicate functions (e.g. `read1`, `readinto1`)? i.e. why not call `read` and `readinto` directly?
- what is the purpose of the `readable()` and `seekable()` methods on `NetFile` when these always return True?

bemoody:
All of these methods are part of the Python file API
(https://docs.python.org/3/library/io.html).  The reason for
implementing them is so that the NetFile object can be plugged into
other libraries that expect a standard file object.

In particular, this will work with the soundfile library for reading
FLAC (and other audio formats).

Unfortunately, one function this *doesn't* work for is np.fromfile,
but we can work around that.

tompollard:
thanks for explaining. looks good to me!

PR #321, opened 2021-08-22
Remove dependency on scikit-learn

hofaflo:
As far as I can tell, `scikit-learn` is only used in the module [processing.qrs](https://github.com/MIT-LCP/wfdb-python/blob/408ad93324c6c2f8c41aaaf51ddc974372cbf837/wfdb/processing/qrs.py#L6). The two uses of the imported function [`normalize`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.normalize.html) could be handled by a [simple implementation in `wfdb`](https://github.com/hofaflo/wfdb-python/blob/0e237dfc96e85a9ef6917a049517f4b7adb563b5/wfdb/processing/basic.py#L235-L250). This would allow `wfdb` to remove `scikit-learn` as a requirement and speed up the import time of `wfdb.processing`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,8 +29,7 @@ class Mock(MagicMock):
     def __getattr__(cls, name):
             return MagicMock()
 
-MOCK_MODULES = ['numpy', 'matplotlib', 'matplotlib.pyplot', 'pandas', 'scipy',
-                'sklearn', 'sklearn.preprocessing']
+MOCK_MODULES = ['numpy', 'matplotlib', 'matplotlib.pyplot', 'pandas', 'scipy']
 sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
 
 # -- General configuration ------------------------------------------------
@@ -177,6 +176,3 @@ texinfo_documents = [
      author, 'wfdb', 'One line description of project.',
      'Miscellaneous'),
 ]
-
-
-
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,6 @@ pyparsing==2.4.7
 python-dateutil==2.8.1
 pytz==2020.1
 requests==2.23.0
-scikit-learn==0.22.2.post1
 scipy==1.4.1
 threadpoolctl==2.1.0
 urllib3==1.25.9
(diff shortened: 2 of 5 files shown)


bemoody:
Thanks; this looks good to me.

It looks to me like the *existing* code in `XQRS._is_twave` was buggy (wfdb/processing/qrs.py, lines 531-545).  If I understand correctly, `scipy.preprocessing.normalize` would return an Nx1 array, `np.diff` would return an Nx0 array, the `<` comparison would return an empty one-dimensional array, so the function would always return false.

I would assume that the new version is in fact what @cx1111 originally intended, and hopefully should improve the overall specificity of the algorithm.  But, Chen, do you think this is okay or should be tested more?

bemoody:
For the old version of XQRS, over mitdb:

- Gross QRS sensitivity: 99.35
- Gross QRS positive predictivity: 99.74
- Average QRS sensitivity: 99.34
- Average QRS positive predictivity: 99.71

For the new version:

- Gross QRS sensitivity: 99.03
- Gross QRS positive predictivity: 99.76
- Average QRS sensitivity: 99.08
- Average QRS positive predictivity: 99.73

(Total false positives decreased from 237 to 216, false negatives increased from 592 to 888.)

Not great but not terrible, and might be useful sometimes.

The old behavior is equivalent to setting the config parameter `t_inspect_period` to zero:
```
qrs_inds = xqrs_detect(sig, fs, conf=XQRS.Conf(t_inspect_period=0))
```
So if we wanted to keep the old behavior by default we could just set the default `t_inspect_period` to 0.

tompollard:
Thanks @hofaflo. Nice checks @bemoody and I think it would make sense to set the default as you suggest. Please could someone take a look at doing this? Once the default is set, it sounds like we should be good to merge.

hofaflo:
Thanks for the benchmark @bemoody!
Done @tompollard.

tompollard:
thanks @hofaflo, looks good to me! okay to merge @bemoody?

bemoody:
Looks good, thanks!

tompollard:
Thanks!

PR #323, opened 2021-08-27
Correctly infer length of record when unspecified

bemoody:
If the header file does not specify the signal length, `rdrecord` tries to determine it based on the length of the first signal file.

This calculation was wrong when the signal file contained a prolog, or any signal contained multiple samples per frame.


tompollard:
Thanks @bemoody, looks good to me.

PR #324, opened 2021-09-01
Allow hyphens in segment names, and fix basic parsing issues

bemoody:
* Hyphens should be permitted in segment names.  Currently hyphens are permitted in record names and signal file names, but not in segment names.

* When parsing a line in a header file, we must start at the beginning of the line.  Currently we search for the expected pattern *anywhere within the line*, which means any "junk" at the start of the line is silently ignored.

This means that if a multi-segment record includes a segment with a hyphen in the name, everything up to and including the hyphen is silently ignored.

A related problem occurs if there are slashes in a segment or signal file name.  Although subdirectories are permitted in WFDB format, wfdb-python currently doesn't support them, so these should raise an exception rather than being silently ignored.

### DIFF
--- a/wfdb/io/_header.py
+++ b/wfdb/io/_header.py
@@ -882,6 +882,8 @@ def _parse_record_line(record_line):
 
     # Read string fields from record line
     match = _rx_record.match(record_line)
+    if match is None:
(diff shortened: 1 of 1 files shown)


tompollard:
thanks Benjamin

PR #328, opened 2021-09-29
Handle sample numbers > 2**31 in annotation files

bemoody:
In an Annotation object, the event timestamps are stored as an array
of integers.  (The array is called `sample` because one "tick" usually
corresponds to one sample in the signal file, though this is not
necessarily the case.)

It's uncommon, but by no means unheard of, to have a record longer
than 2147483647 ticks (for example, the vendor annotations for
MIMIC-IV waveforms have 1-millisecond resolution, and many of those
records are longer than 24 days.)  Therefore, we must be sure that the
`sample` array is able to accommodate larger values, which means the
timestamps must be calculated and stored as 64-bit integers, rather
than numpy's default integer type.

Furthermore, timestamps in WFDB annotation files are stored as an
*offset* from the previous annotation.  If the offset is between 0 and
1023, then the offset along with the annotation type is stored as two
bytes.  If the offset is negative or larger than 1023, then a "SKIP"
sequence is used to specify a 32-bit signed offset.

If the offset between two annotations is larger than 2147483647 (or
less than -2147483648), then multiple SKIPs must be used in a row.
Both `rdann` and `wrann` need to accommodate this case.


bemoody:
Well, that's crazy:
```
2021-09-29T17:26:28.3441563Z ##[group]Run nosetests
2021-09-29T17:26:28.3442229Z ESC[36;1mnosetestsESC[0m
2021-09-29T17:26:28.3491355Z shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
2021-09-29T17:26:28.3491907Z env:
2021-09-29T17:26:28.3492446Z   pythonLocation: C:\hostedtoolcache\windows\Python\3.6.8\x64
2021-09-29T17:26:28.3493080Z ##[endgroup]
2021-09-29T17:26:33.8184778Z ...D:\a\wfdb-python\wfdb-python\wfdb\io\annotation.py:1883: RuntimeWarning: overflow encountered in long_scalars
2021-09-29T17:26:33.8186197Z   sample_diff += skip_diff
2021-09-29T17:27:41.5789858Z F.........................................
2021-09-29T17:27:41.5790791Z ======================================================================
2021-09-29T17:27:41.5791673Z FAIL: Read and write annotations with large time skips
2021-09-29T17:27:41.5792556Z ----------------------------------------------------------------------
2021-09-29T17:27:41.5793298Z Traceback (most recent call last):
2021-09-29T17:27:41.5795825Z   File "D:\a\wfdb-python\wfdb-python\tests\test_annotation.py", line 198, in test_4
2021-09-29T17:27:41.5797271Z     self.assertEqual(annotation.sample[0], 10000000000)
2021-09-29T17:27:41.5798112Z AssertionError: 1410065408 != 10000000000
2021-09-29T17:27:41.5798595Z 
2021-09-29T17:27:41.5799226Z ----------------------------------------------------------------------
2021-09-29T17:27:41.5799952Z Ran 45 tests in 72.082s
2021-09-29T17:27:41.5800367Z 
2021-09-29T17:27:41.5800921Z FAILED (failures=1)
2021-09-29T17:27:41.9665736Z ##[error]Process completed with exit code 1.
```

Clearly something is trying to stuff timestamps into 32-bit integers (which is wrong, of course).  The obvious culprit is this:
```
def lists_to_int_arrays(*args):
    return [np.array(a, dtype='int') for a in args]
```
but for me, that raises an OverflowError if the value is too large, it doesn't merely show a warning.

tompollard:
Thanks Benjamin. Chen, it looks like your points are resolved, so I'm taking the liberty of merging!

PR #329, opened 2021-09-30
Test suite improvements

bemoody:
1. While running the test suite, set the numpy error handling mode to raise exceptions on arithmetic errors (as opposed to the default behavior of showing a warning, which doesn't cause the test to fail and is difficult to debug.)

2. In addition to the current matrix of tests, run tests in a 32-bit Debian 10 (oldstable) environment, in order to catch any errors that might result from different default integer sizes (is numpy.dtype(int) an int32 or an int64?) or different FP instruction sets, and also to check backward compatibility with stable versions of numpy/scipy/pandas/matplotlib/requests.


bemoody:
Regarding the i386 tests: GitHub apparently has no problem running i386 containers with docker; however, the `actions/checkout@v2` command requires amd64 libc6 and libstdc++6 to be installed in the container.  (https://github.com/actions/checkout/issues/334).  `actions/checkout@v1` doesn't have this issue, though it is obviously lacking some newer features.

Of course it's possible to install amd64 libraries using `dpkg --add-architecture amd64 && apt-get update && apt-get install libstdc++6:amd64`, after which `actions/checkout@v2` will work.

But for purposes of this test I actually prefer not to have any amd64 libraries installed, to be sure we're not inadvertently relying on them and to keep things simpler and faster.  So I've used `actions/checkout@v1` here, and hopefully that will keep working and/or GitHub will eventually figure out how to fix the above issue.

Issue #306, opened 2021-06-15
The `dl_files` function is giving an error


briangow:
The `wfdb.io.dl_files` function is working in older version of python (v3.6) but is failing with the following error in a newer version (3.9.5):

```
Downloading files...
Traceback (most recent call last):
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/multiprocessing/spawn.py", line 125, in _main
    prepare(preparation_data)
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/multiprocessing/spawn.py", line 236, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/multiprocessing/spawn.py", line 287, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/runpy.py", line 268, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/temp_credentialed_download.py", line 5, in <module>
    wfdb.io.dl_files('mitdb','/Users/briangow/Documents/MIT/physionet/files',['115.dat'])
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/site-packages/wfdb/io/download.py", line 531, in dl_files
    pool = multiprocessing.Pool(processes=2)
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/multiprocessing/context.py", line 119, in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild,
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/multiprocessing/pool.py", line 212, in __init__
    self._repopulate_pool()
  File "/Users/briangow/Documents/MIT/physionet/physionet-build/env/lib/python3.9/multiprocessing/pool.py", line 303, in _repopulate_pool
[...]

PR #330, opened 2021-10-01
Fix downloading files in windows

lbugnon:
Adding a small change to make optional the use of multiple threads for download files. We couldn't make it work in a conda environment in windows.

With this change, default is the same, but multithread download can be disabled by optional parameter.

### DIFF
--- a/wfdb/io/download.py
+++ b/wfdb/io/download.py
@@ -262,7 +262,7 @@ def get_record_list(db_dir, records='all'):
 
     """
     # Full url PhysioNet database
-    if '/' not in db_dir:
+    if os.sep not in db_dir:
         db_url = posixpath.join(config.db_index_url, db_dir, record.get_version(db_dir))
     else:
         db_url = posixpath.join(config.db_index_url, db_dir)
--- a/wfdb/io/download.py
+++ b/wfdb/io/download.py
@@ -511,10 +511,6 @@ def dl_files(db, dl_dir, files, keep_subdirs=True, overwrite=False):
     make_local_dirs(dl_dir, dl_inputs, keep_subdirs)
 
     print('Downloading files...')
-    # Create multiple processes to download files.
-    # Limit to 2 connections to avoid overloading the server
-    #pool = multiprocessing.Pool(processes=2)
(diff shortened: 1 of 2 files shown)


bemoody:
Thanks!

As I said before, did you try simply changing "multiprocessing.Pool" to
"multiprocessing.dummy.Pool"?  I know that multiprocessing is messy and
best avoided here, but multiprocessing.dummy should not have the same
drawbacks (it has other drawbacks instead.)

If multiprocessing.dummy.Pool doesn't work for you either, that's a
bigger problem and something I'd like to try to understand.

cx1111:
I think we can merge in this change after the requested changes are made. It seems useful in general.

Let's figure out #306 separately.

cx1111:
I think we can merge in this change after the requested changes are made. It seems useful in general.

Let's figure out #306 separately.

bemoody:
If we do want to make parallel downloads optional, I'd say the name of the argument should be `parallel` or something like that.  If we change the implementation to use threads instead of processes then `use_multiprocess` becomes misleading.

I'm not opposed to adding that option, I just don't want to add the option *solely to avoid a known bug in the package*, instead of just fixing the bug.

I also don't understand this change, which seems unrelated:
```
@@ -4525,7 +4527,8 @@ def dl_database(db_dir, dl_dir, records='all', annotators='all',
         db_dir = posixpath.join(dir_list[0], get_version(dir_list[0]), *dir_list[1:])
     else:
         db_dir = posixpath.join(db_dir, get_version(db_dir))
-    db_url = posixpath.join(download.PN_CONTENT_URL, db_dir) + '/'
+
+    db_url = posixpath.join(download.PN_CONTENT_URL, db_dir)
     # Check if the database is valid
     _url.openurl(db_url, check_access=True)
 
```
I think this is probably a mistake, we *do* want to request a path ending in slash (the path with the slash is the "real" URL, without the slash is a redirection.)  I don't remember off the top of my head if check_access=True will treat a redirected URL as valid or not.

cx1111:
Yes, let's set the param name to`parallel`.

And I agree that this change isn't meant as a way to avoid solving the actual issue.

PR #331, opened 2021-10-13
Read multi-frequency multi-segment records

bemoody:
A multi-segment record can contain signals with multiple sampling frequencies, as in the MIMIC DB or the upcoming MIMIC-IV Waveform DB.

By default, `rdrecord` will convert the input to a uniform 2D array of samples; in the case of multi-frequency records it downsamples all signals to the frame frequency ("smooth frames"), and in the case of multi-segment records it stitches the input arrays together into one ("multi-to-single").  However, both these options can be turned off.

In particular, smooth frames mode is simply not a good idea; when actually analyzing a multi-frequency record we need to be able to read each signal at its original sampling frequency or better.

Currently, `rdrecord` disallows this:
```
>>> r = wfdb.rdrecord('041', pn_dir='mimicdb/041', sampto=10, smooth_frames=False)
...
ValueError: This package version cannot expand all samples when reading multi-segment records. Must enable frame smoothing.
```

It also fails in "non-multi-to-single" mode:
```
>>> r = wfdb.rdrecord('041', pn_dir='mimicdb/041', sampto=10, smooth_frames=False, m2s=False)
...
ValueError: This package version cannot expand all samples when reading multi-segment records. Must enable frame smoothing.
```

These changes fix both cases.

This is based on and depends on pull #313 (it might possibly work for mimicdb without those changes, FSVO "work", but both sets of changes are definitely needed for MIMIC-IV.)


wfdb/io/record.py
```diff
@@ -1152,7 +1162,16 @@ def multi_to_single(self, physical, return_res=64):
                             reference_fields[field][ch] = item_ch
                         # mismatch case
                         elif reference_fields[field][ch] != item_ch:
-                            if physical:
+                            if field == 'samps_per_frame':
+                                raise ValueError(
```
cx1111, line 1166: Use fstrings please.

tests/test_record.py
```diff
@@ -630,6 +630,34 @@ def test_multi_fixed_c(self):
...
+
+        sig_target = np.genfromtxt(
+            'tests/target-output/record-multi-fixed-d.gz')
+
+        record_named = wfdb.rdrecord('sample-data/multi-segment/041s/041s',
+                                     channel_names=['ABP', 'V', 'I'],
+                                     physical=False, smooth_frames=False)
+
+        np.testing.assert_array_equal(sig, sig_target)
+        assert record.__eq__(record_named)
```
cx1111, line 659: Does `assert record == record_named` work? Also, please add a comment above each rdrecord call in this test to explain the intention/difference.

cx1111 (approved):
Look good, thanks for the new test cases. Please apply formatting before merge.

We really should rename the tests so that we can more easily see the various scenarios being covered.

bemoody:
> Does `assert record == record_named` work?

It certainly should and is more readable.  I'd go further than that and
use self.assertEqual instead.

The Record.__eq__ method is a bit messy though for two reasons: the
verbose flag and the use of np.testing.assert_array_equal.  Both are
useful for testing but break the standard Python API.  Calling
__eq__ really shouldn't have any side effects, and should just return a
boolean to say if the two objects are equal.  It's a mess and partly I
blame numpy for making it a mess.

So what I actually think would be better is to define a method
Record.assert_equal that does the verbose thing.  Possibly this can be
hooked into unittest via addTypeEqualityFunc.

> Also, please add a comment above each rdrecord call in this test to
> explain the intention/difference.

Will do.  Thanks.

bemoody:
Rebased to apply black formatting.  Here's how:
```
rm -f .git/info/attributes
git checkout multisegment-multifrequency
git rebase 1e695017fa4722c8008e02ad514086ca8eb77d76
git config --local filter.black.clean 'black -tpy37 -l80 -q -'
git config --local filter.black.smudge 'black -tpy37 -l80 -q -'
echo '*.py filter=black' > .git/info/attributes
git rebase -Xrenormalize 3b408f23f50889c7d1c6fee9329c76b739b9f667
rm -f .git/info/attributes
```
Commit d25923b2b617d52ba6b68885c3e8138fb0653fdd became a no-op and was therefore automatically dropped.

PR #342, opened 2022-03-01
Eliminate "is True" and "is False"

bemoody:
If a parameter X is documented as having a boolean value, it should be treated (in duck-typing style) as true if bool(X) is true, or false if bool(X) is false.

Writing "if X is True" is almost always a bad idea: it suggests to the reader that "X is True" and "X is False" are the only possibilities, when in fact there is no such guarantee.

(In some cases it might make sense to explicitly check argument types to catch programming errors - but in most cases this feels like an overcomplication, and *even then* it's a bad idea to write "if X is True" because the permitted argument types might change in the future.)

To wit: `rdrecord("foo", physical=0)` would do what you expect; `rdrecord("foo", physical=1)` would crash.  `rdrecord("foo", smooth_frames=1)` would do what you expect; `rdrecord("foo", smooth_frames=0)` would crash.

### DIFF
--- a/wfdb/io/_signal.py
+++ b/wfdb/io/_signal.py
@@ -1891,7 +1891,7 @@ def _np_dtype(bit_res, discrete):
         if bit_res <= np_res:
             break
 
-    if discrete is True:
+    if discrete:
         return 'int' + str(np_res)
     else:
         # No float8 dtype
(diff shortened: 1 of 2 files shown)


tompollard:
looks good, thanks!

PR #343, opened 2022-03-11
Record.convert_dtype should fail if expected data is missing

bemoody:
This is not directly related to other issues surrounding `smooth_frames`, but a minor bug that probably doesn't affect anyone at present.

`Record.convert_dtype` can be used to change the storage format of the signal array(s).  (It's unlikely that applications will ever use this function directly and it's not included in the package documentation.)

However, if `convert_dtype` is called with `physical=True` and `smooth_frames=False`, it should convert the data type(s) of `e_p_signal`.  It shouldn't try to modify `p_signal` in this case (and the previous code would not have worked correctly either.)  If you ask to convert the data type of `e_p_signal`, and `e_p_signal` is None, this should raise an exception.

### DIFF
--- a/wfdb/io/_signal.py
+++ b/wfdb/io/_signal.py
@@ -705,14 +705,9 @@ class SignalMixin(object):
                 if current_dtype != return_dtype:
                     self.p_signal = self.p_signal.astype(return_dtype, copy=False)
             else:
-                if self.e_p_signal is not None:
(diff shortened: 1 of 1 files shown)


bemoody:
(pull #244 is where this was introduced)

tompollard:
Looks good to me, thanks.

PR #345, opened 2022-03-17
Show annotation markers atop signals

bemoody:
When plotting signals and annotations using `wfdb.plot_items` or `wfdb.plot_wfdb
`, the annotation markers are currently displayed "underneath" the signal trace, where they are hard to see.  Put them "on top" instead.

For example, compare the appearance of running `wfdb.plot_wfdb(wfdb.rdrecord('sample-data/100', sampto=1000), wfdb.rdann('sample-data/100', 'qrs', sampto=1000))` before and after this change.


tompollard:
Agreed, looks much better. 

Pre-change:

image

Post-change:

image

PR #346, opened 2022-03-18
Make wfdb.rdann, wfdb.rdrecord, wfdb.rdsamp also accept pathlib.Path objects as the input record_name

wenh06:
Currently, the functions `wfdb.rdann`, `wfdb.rdrecord`, `wfdb.rdsamp` only accept `str` for the input `record_name`, as they have `str` specific operations. `wfdb.rdheader` currently accepts `pathlib.Path` objects as input `record_name` since the `os.path` module is able to handle `pathlib.Path` as well as `str` type input.

For example, on a linux machine
```python
import pathlib, wfdb
sp = "/home/wenhao/Jupyter/wenhao/data/CPSC2021/training_I/data_0_1"
pp = pathlib.Path(sp)
wfdb.rdrecord(pp)
```
would raise error
```python
AttributeError                            Traceback (most recent call last)
<ipython-input-6-754a0c10c458> in <module>
----> 1 wfdb.rdrecord(pp)

~/.local/lib/python3.6/site-packages/wfdb/io/record.py in rdrecord(record_name, sampfrom, sampto, channels, physical, pn_dir, m2s, smooth_frames, ignore_skew, return_res, force_channels, channel_names, warn_empty)
   3426         pn_dir = posixpath.join(dir_list[0], get_version(dir_list[0]), *dir_list[1:])
   3427 
-> 3428     if record_name.endswith('.edf'):
   3429         record = edf2mit(record_name, pn_dir=pn_dir, record_only=True)
   3430     elif record_name.endswith('.wav'):

AttributeError: 'PosixPath' object has no attribute 'endswith'
```
Sometimes one would like to work locally with `pathlib`, which was introduced in Python 3.4 ([PEP 428](https://www.python.org/dev/peps/pep-0428/)), to better handle the file paths.

### DIFF
--- a/wfdb/io/annotation.py
+++ b/wfdb/io/annotation.py
@@ -4,6 +4,7 @@ import os
 import pandas as pd
 import re
 import posixpath
+import pathlib
 import pdb
 import struct
 import sys
@@ -146,7 +147,7 @@ class Annotation(object):
                  label_store=None, description=None, custom_labels=None,
                  contained_labels=None):
 
-        self.record_name = record_name
+        self.record_name = str(record_name)
         self.extension = extension
 
         self.sample = sample
@@ -1584,7 +1585,7 @@ def rdann(record_name, extension, sampfrom=0, sampto=None, shift_samps=False,
 
     Parameters
     ----------
-    record_name : str
+    record_name : str or pathlib.Path
         The record name of the WFDB annotation file. ie. for file '100.atr',
         record_name='100'.
     extension : str
@@ -1678,7 +1679,7 @@ def rdann(record_name, extension, sampfrom=0, sampto=None, shift_samps=False,
             pass
 
     # Create the annotation object
-    annotation = Annotation(record_name=os.path.split(record_name)[1],
+    annotation = Annotation(record_name=pathlib.Path(record_name).name,
                             extension=extension, sample=sample,
                             label_store=label_store,  subtype=subtype,
                             chan=chan, num=num, aux_note=aux_note, fs=fs,
@@ -1747,7 +1748,7 @@ def load_byte_pairs(record_name, extension, pn_dir):
 
     Parameters
     ----------
-    record_name : str
+    record_name : str or pathlib.Path
         The record name of the WFDB annotation file. ie. for file '100.atr',
         record_name='100'.
     extension : str
@@ -1764,13 +1765,14 @@ def load_byte_pairs(record_name, extension, pn_dir):
         The input filestream converted to an Nx2 array of unsigned bytes.
 
     """
+    file_name = pathlib.Path(record_name).with_suffix('.' + extension)
     # local file
     if pn_dir is None:
-        with open(record_name + '.' + extension, 'rb') as f:
+        with open(file_name, 'rb') as f:
             filebytes = np.fromfile(f, '<u1').reshape([-1, 2])
     # PhysioNet file
     else:
-        filebytes = download._stream_annotation(record_name+'.'+extension, pn_dir).reshape([-1, 2])
+        filebytes = download._stream_annotation(file_name, pn_dir).reshape([-1, 2])
 
     return filebytes
 
@@ -2697,7 +2699,7 @@ def rdedfann(record_name, pn_dir=None, delete_file=True, info_only=True,
         })
         df_out = _format_ann_from_df(df_in)
         # Remove extension from input file name
-        record_name = record_name.split(os.sep)[-1].split('.')[0]
+        record_name = pathlib.Path(record_name).stem
         extension = 'atr'
         fs = rec.fs
         sample = (df_out['onset'].to_numpy()*fs).astype(np.int64)
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -2,6 +2,7 @@ import datetime
 import multiprocessing
 import posixpath
 import re
+import pathlib
 
 import numpy as np
 import os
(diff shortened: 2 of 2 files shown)


bemoody:
Thank you!  This looks like a great improvement.

I'm not entirely familiar with `pathlib` so I'm a little unsure of some of the details.

One thing to be wary of is that "file names" or "record names" might be either filesystem paths or URL paths.  Some existing functions permit `record_name` to include a subdirectory when streaming data from PhysioNet; other functions don't allow that, and silently ignore any directory name in `record_name` (which I consider a bug.)

Here, for example, it looks like you're passing a `pathlib.Path` object to `download._stream_annotation`.  Is that a good idea or not?  I guess that `posixpath.join` will automatically convert the argument back into a string, but will it correctly use slashes if running on a Windows platform?  And does that mean that rdann would now accept backslashes in URLs, only on Windows?

If we do want the `download` functions to accept a `pathlib.Path` as an alternative to `str` then that should be documented.

I'm not necessarily opposed to either saying "`pathlib.Path` may be used only for local files" or saying "`pathlib.Path` may be used either for local files or for URLs", but I do think it's important to handle URL arguments in the *same way* on all hosts.

wenh06:
> Thank you! This looks like a great improvement.
> 
> I'm not entirely familiar with `pathlib` so I'm a little unsure of some of the details.
> 
> One thing to be wary of is that "file names" or "record names" might be either filesystem paths or URL paths. Some existing functions permit `record_name` to include a subdirectory when streaming data from PhysioNet; other functions don't allow that, and silently ignore any directory name in `record_name` (which I consider a bug.)
> 
> Here, for example, it looks like you're passing a `pathlib.Path` object to `download._stream_annotation`. Is that a good idea or not? I guess that `posixpath.join` will automatically convert the argument back into a string, but will it correctly use slashes if running on a Windows platform? And does that mean that rdann would now accept backslashes in URLs, only on Windows?
> 
> If we do want the `download` functions to accept a `pathlib.Path` as an alternative to `str` then that should be documented.
> 
> I'm not necessarily opposed to either saying "`pathlib.Path` may be used only for local files" or saying "`pathlib.Path` may be used either for local files or for URLs", but I do think it's important to handle URL arguments in the _same way_ on all hosts.

You're correct. On a Windows system, if you pass a `Path` object (actually a `WindowsPath` object) as the `file_name`, then error might occur when this `file_name` includes a subdirectory for
```python
url = posixpath.join(config.db_index_url, pn_dir, file_name)
```
in the function `download._stream_annotation`. Such error can be avoided by force converting `file_name` to a `PosixPath` using `pathlib.PurePosixPath` as
```python
url = posixpath.join(config.db_index_url, pn_dir, pathlib.PurePosixPath(file_name))
```
or using
```python
url = config.db_index_url / pathlib.PurePosixPath(pn_dir / file_name)
```
given `file_name` is a `Path` object. A drawback of `pathlib` is that it cannot handle the URLs correctly (mainly the double slashes), for example
```python
... pathlib.PurePosixPath('https://physionet.org/files/')
>>> PurePosixPath('https:/physionet.org/files')
```

The following image is an example tested on a Windows machine
image

wenh06:
> One thing to be wary of is that "file names" or "record names" might be either filesystem paths or URL paths. Some existing functions permit record_name to include a subdirectory when streaming data from PhysioNet; other functions don't allow that, and silently ignore any directory name in record_name (which I consider a bug.)

I also think this is a bug:
```python
>>> wfdb.rdrecord("x_mitdb/x_111", pn_dir="mitdb", sampto=2000)
NetFileNotFoundError: 404 Error: Not Found for url: https://physionet.org/files/mitdb/1.0.0/x_111.hea
>>> wfdb.rdrecord("x_111", pn_dir="mitdb/1.0.0/x_mitdb/", sampto=2000)
<wfdb.io.record.Record at 0x7f6cd4ee7c50>
```

wenh06:
I think there's a lot of work to do to better deal with the paths. For example, I found currently `wfdb` detects whether a `pn_dir` contains the version number using
```python
'/' not in db_dir
```
or
```python
'.' not in pn_dir
```
I think it's better to specify a pattern for the database version and use `re.search` to detect it
```python
DB_VERSION_PATTERN = re.compile("\d+.\d+\.\d+")
```

cx1111:
Thanks very much for identifying the bugs. We're going to think a bit more about how we want the API to be in the next major version, such that the behavior is more explicit. Especially when adding cloud data sources.

wfdb/io/annotation.py
```diff
@@ -1764,13 +1765,14 @@ def load_byte_pairs(record_name, extension, pn_dir):
         The input filestream converted to an Nx2 array of unsigned bytes.
 
     """
+    file_name = pathlib.Path(record_name).with_suffix('.' + extension)
     # local file
     if pn_dir is None:
-        with open(record_name + '.' + extension, 'rb') as f:
+        with open(file_name, 'rb') as f:
```
Ivorforce, line 1771: paths offer a `.open` function, too, which might be more readable than using the global one.

wfdb/io/record.py
```diff
@@ -4373,7 +4377,7 @@ def wrsamp(record_name, fs, units, sig_name, p_signal=None, d_signal=None,
 
     """
     # Check for valid record name
-    if '.' in record_name: 
+    if '.' in str(record_name):
```
Ivorforce, line 4380: As directories might contain dots, maybe this wants to check for record_name.name instead.

Ivorforce (commented):
Would love to see this merged!

wenh06:
> Thanks very much for identifying the bugs. We're going to think a bit more about how we want the API to be in the next major version, such that the behavior is more explicit. Especially when adding cloud data sources.

I agree with you. I tried continuing to correct (enhance) path operations in other places for several hours but finally gave up. I hope that path operations would be designed in a more uniform and modern way in the next major version.

PR #356, opened 2022-04-12
Switch to Poetry for dependency management and package description

cx1111:
Time to switch to the newer tool used by many projects including scipy and fastapi.

https://python-poetry.org/docs/basic-usage/

Benefits:
- Makes it easier to specify different groups of dependencies. eg. Dev, plotting, processing, etc.
- Requires fewer different setup files and packages. Replaces: `setup.py`, `MANIFEST.in`, `requirements.txt`, `version.py` (to be discussed) and `twine`.
- Gives a warning/error when dependency requirements clash.
- Does not require manually keeping `requirements.txt` in sync with `setup.py`. When you run `poetry add <package>`, it'll update both `pyproject.toml` and `poetry.lock`.

This is also in preparation of adding more dev dependencies: pylint and blank.

Speaking of the dependency clash, this PR also updates a few project dependencies. Most notable are:
- Python versions. Now supported Python versions are ~3.8~ 3.7-3.10. ~Python 3.7 is already near end of life~: https://www.python.org/downloads/
- Bump scipy to 1.7.0
- Bump numpy to 1.22.3. There's currently a dependabot PR telling us to upgrade our numpy.

Specifying dep versions: https://python-poetry.org/docs/dependency-specification/


pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[tool.poetry]
+name = "wfdb"
+version = "4.0.0-alpha0"
+description = "The WFDB Python package: tools for reading, writing, and processing physiologic signals and annotations."
+authors = ["The Laboratory for Computational Physiology <support@physionet.org>"]
```
tompollard, line 5: Does support@ exist? Maybe use contact@physionet.org instead?

pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[tool.poetry]
+name = "wfdb"
+version = "4.0.0-alpha0"
```
tompollard, line 3: Are you intentionally bumping the version here?

tompollard (approved):
Looks good to me. We've started to (partly) use Poetry for https://github.com/MIT-LCP/physionet-build so we have some familiarity with it. 

I've added a couple of minor comments (+ tests look like they are failing, though haven't checked why).

pyproject.toml
cx1111: Yes, I want to make sure this works if I upload it to pypi. Hence, the alpha version.

The way this package is installed is going to be different (users will explicitly have to specify processing/plot to install those deps), so it will be a new major version (coupled with other changes).

tompollard:
I assume you have to specify the python version for the ubuntu-latest build to get the tests to pass (currently it looks like it wants to install Python 3.1). @bemoody set up this test so we should double check he's okay with dropping the older Python versions here.

bemoody:
I'm very much in favor of adding test cases for newer systems and newer external packages.  I'm very much opposed to dropping test cases or breaking compatibility for no reason.

This package shouldn't be in the business of requiring or even recommending specific external package versions.  You should use whatever version of numpy is available for your system, and this package should be *compatible* with as wide a range of numpy versions as is reasonably possible.

Removing `wfdb.__version__` is a major API break ("super-major", even?) and not to be done lightly.

Removing setup.py sounds like a pretty bad idea too.

cx1111:
Restored the Python 3.7 version compatibility. Found a set of dependencies which work for Python 3.7-3.10, which also include scipy 1.7 for M1 macs: https://github.com/scipy/scipy/issues/13409

cx1111:
> I'm very much in favor of adding test cases for newer systems and newer external packages. I'm very much opposed to dropping test cases or breaking compatibility for no reason.
> 
> This package shouldn't be in the business of requiring or even recommending specific external package versions. You should use whatever version of numpy is available for your system, and this package should be _compatible_ with as wide a range of numpy versions as is reasonably possible.

So overall, I agree that we should try to support as many different platforms as possible, as long as they do not impair the development of this package. Removing P37 was a mistake. However, there are several important points to be considered:

This package has always had dependencies, as specified dependencies, as specified in`setup.py`. It does not work without these dependencies. The current system also specified a compatible version range for each package.

Poetry has forced the package to explicitly set a more restrictive dependency (independent of me increasing the dependency versions), by highlighting the restrictions which were always there. They were previously not shown because we had no way of testing them. Adding a larger testing matrix and running our sparse unit tests (which is very useful and I fully support independenly,) in order to figure out compatibility is vastly inferior to having poetry deterministically figuring out compatibility. If our dependencies all have their own dependency graphs and limitations, and we choose to ignore them, as long as our unit tests pass, we're not doing our users any favors. These changes are just making the dependencies explicit, where they were previously hidden. I see this similar to catching errors at compile-time rather than run-time.

Finally, I'll say that there is a tradeoff in keeping a wider compatibility matrix. An exaggerated example would be to support Python 2 and 3. There's much overhead in terms of development time, which at some point, just isn't worth it. Especially because there are so few of us active contributors, we need to be careful about how much time we dedicate to this.

> Removing `wfdb.__version__` is a major API break ("super-major", even?) and not to be done lightly.

Restored. I wanted to rely on another method for getting the version info but that's only supported for P3.8.

> Removing setup.py sounds like a pretty bad idea too.

It's redundant with poetry and `pyproject.toml`, which already gives the ability to:
- Build distributions
- Upload to pypi
- Specify metadata and dependencies
- Preserves the ability to pip install

I don't see a reason to keep it. See, for example, the [fastapi repo](https://github.com/tiangolo/fastapi/), which doesn't have one. I imagine you're not against using poetry itself. If so, that's an entirely different discussion.

bemoody:
> It's redundant with poetry and `pyproject.toml`

I was confused by this: apparently pip can use the information in pyproject.toml as a substitute for setup.py.  That seems quite reasonable.  I don't understand why pyproject.toml mentions "tool.poetry" if this file is not related to poetry, but that's okay.

bemoody:
> Poetry has forced the package to explicitly set a more restrictive dependency (independent of me increasing the dependency versions), by highlighting the restrictions which were always there.

I haven't used poetry so I don't know what it does or doesn't force you to do.  If it forces you to make your own software less forwards- or backwards-compatible then that's terrible design and a good reason not to use it.  But I doubt it actually forces you to do that.

I'll be specific: wfdb-python is known to work with these:
- numpy 1.16.2
- pandas 0.23.3
- requests 2.21.0
- scipy 1.1.0
- matplotlib 3.0.2
and if any of these do break the package, I'd be inclined to fix it.

I don't see any reason to suspect that wfdb-python will *not* work with these, once they are released:
- Python 3.11
- pandas 1.4.0
- requests 2.28.0
- scipy 1.8.0
- matplotlib 3.6.0
and if any of these do break the package, then it certainly needs to be fixed.

Preemptively declaring that any unsupported/untested version is incompatible is a nightmare, both for people in the future and for people in the present who want to use your library in combination with *anything else*.

A separate question to what external package versions are compatible is whether you want to ship, in this repository, a massive list of hashes for other packages.  I think that's not a great idea because that list will always be incomplete and out of date.  *I'm* not inclined to go to any effort to maintain such a list and certainly not to test or audit those packages.  But I suppose you can argue the list doesn't hurt if people aren't required to use it.

bemoody:
Can you clarify what you mean by "deterministically figuring out compatibility"?  Maybe there is some point there I have missed.

If there is some automated mechanism that's telling you *this package isn't compatible with numpy 1.16*, that would be really interesting (even if I don't believe it.)  If it's telling you *numpy 1.16 isn't officially supported or officially developed*, that's also an interesting piece of information but not relevant in any technical sense.

(If it's telling you *it's not possible to prove, on the basis of what you currently have installed, that numpy 1.16 is compatible*, that's very different from telling you *numpy 1.16 is incompatible.*)

If someone is telling you that *this package is not going to be compatible with requests 2.28*, their crystal ball needs adjusting.

cx1111:
> Can you clarify what you mean by "deterministically figuring out compatibility"? Maybe there is some point there I have missed.
> 
> If there is some automated mechanism that's telling you _this package isn't compatible with numpy 1.16_, that would be really interesting (even if I don't believe it.) If it's telling you _numpy 1.16 isn't officially supported or officially developed_, that's also an interesting piece of information but not relevant in any technical sense.
> 
> (If it's telling you _it's not possible to prove, on the basis of what you currently have installed, that numpy 1.16 is compatible_, that's very different from telling you _numpy 1.16 is incompatible._)
> 
> If someone is telling you that _this package is not going to be compatible with requests 2.28_, their crystal ball needs adjusting.

Yes, there's an automated mechanism that tells you if a set of dependencies you list clashes. Including Python versions. Try it out and see.

What I'll do is revert the dependencies to their existing values, and try to upgrade them in a separate PR so that the two issues are not confounded.

cx1111:
Ok, I've converted all the tilde dependency ranges to use caret for more flexibility.

All dependency versions in `pyproject.toml` now have minimum versions matching what's currently set in `setup.py`. The exceptions are scipy and pandas which are set to 1.0.0, since they are probably due for a major version upgrade, given that we're going to support one major version per dependency.

Python is ^3.6.1 because Pandas will not allow it to be ^3.6

Let me know if this dependency set is acceptable. I'm going to try to fix the actions CI.

bemoody:
Yeah that makes sense.  Thanks!

cx1111:
It looks like according to: https://github.com/MIT-LCP/wfdb-python/actions/runs/2185363249 that we are having issues with Python 3.6 (which is past its end of life) and 3.10. The 3.10 error only comes from nosetests.

I tried running Pytest on a 3.10 container and it does work. Here's what I suggest:
- Remove 3.6 and 3.10 support and merge this PR.
- Swap nose for pytest in the next PR, and re-add 3.10 support. Thoughts?

.github/workflows/run-tests.yml
```diff
@@ -27,11 +27,12 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements.txt
-      - name: Run nosetests
-        run: |
-          nosetests
+          python -m pip install --upgrade pip poetry
+          pip install ".[dev]"
```
cx1111, line 31: We're using pip to install the dependencies instead of poetry for a few reasons:
- We want to support pip installations. This will probably be the primary way
- A deployment doesn't actually need poetry, which would add another dependency. Use poetry primarily for developing.
- Poetry has its own system of managing environments, and where packages are installed by default, which can get complex. https://realpython.com/dependency-management-python-poetry/ eg. When trying to run the CI earlier with poetry, `poetry install` would install the packages (including the test executable) into `~/.cache` which weren't automatically usable by the system installed Python (causing all those test failures). We don't need to deal with this.

tompollard:
> Remove 3.6 and 3.10 support and merge this PR.
> Swap nose for pytest in the next PR, and re-add 3.10 support. Thoughts?

Both sound fine to me. 

Often you can do a straight swap with pytest. Nosetests is old and should be replaced regardless.

bemoody:
A few things:

- nosetests indeed seems to be unusable on python 3.10.  pytest appears to work fine as a replacement.

- The package passes tests using python 3.10 and packaged libraries on debian sid ("apt-get install python3.10 python3-numpy python3-scipy python3-matplotlib python3-pandas python3-requests python3-pytest").

- The package passes tests using python 3.6 and packaged libraries on ubuntu 18.04 ("apt-get install python3.6 python3-numpy python3-scipy python3-matplotlib python3-pandas python3-requests python3-pytest").

- If you want to use the packaged pytest (python3-pytest) on Debian or Ubuntu, you must run "pytest-3".  "pytest" is python2.

I think it would be fine (as a first step) to limit the github workflows to python 3.7 through 3.9, but pyproject.toml should not declare the package as being incompatible with python 3.10, because it isn't.

I don't have a strong feeling about support for python 3.6, but bear in mind that Ubuntu 18.04 is still in its "general support" period for another year.

PR #357, opened 2022-04-12
Fix installation

bemoody:
Revert the previous commit (pull #355) which broke `pip` and was merged without review and (apparently) without passing tests.


cx1111:
Can we just change this: https://github.com/MIT-LCP/wfdb-python/blob/master/setup.py#L14 to read the markdown file instead?

PR #358, opened 2022-04-12
Remove requirements.txt; expand workflows to include Python 3.9/3.10

bemoody:
Including a "requirements.txt" (or equivalent) in the repository likely suggests to users that those are the specific package versions they need to use.  That's not true.

Indeed, as this package is intended in part as a general-purpose library for applications to use (not just an application in and of itself), it's important that this package *doesn't* impose unnecessary limits on what other libraries may be co-installed with it.

So for testing purposes, it would be better to actually test a variety of *different yet stable* package versions.  'test-deb10-i386' is a start at doing that.  But in lieu of that, we should at least try to test "the latest thing available from pypi" across a variety of interpreter versions.

You may think this makes test-suite runs less reproducible... which is true.  But they weren't really reproducible *anyway* because pip would install a whole bunch of indirect dependencies automatically.  If we need to reproduce a test-suite run, we can do so (or try to) by extracting the list of installed versions from the log.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install .
       - name: Run nosetests
         run: |
           nosetests
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


bemoody:
Github is not showing any logs for this or for pull #357 for some reason.  Presumably, though, the workflows will fail due to pull #355.

bemoody:
```
$ curl https://api.github.com/repos/MIT-LCP/wfdb-python/actions/runs?branch=loose-versions
{
  "total_count": 0,
  "workflow_runs": [

  ]
}
```
*shrug*

cx1111:
Make sure no info is being lost from here, since people have had to manually keep this and setup.py in sync.

bemoody:
These issues are addressed by pulls #356 and #363.

Issue #347, opened 2022-03-18
Partially-completed plots displayed after an error


bemoody:
If you call, say, `plot_wfdb` with some invalid arguments:
```
r = wfdb.rdrecord('sample-data/100', sampto=1000)
wfdb.plot_wfdb(r, ecg_grids=0xdeadbeef)
```
it will raise an exception (`TypeError: 'int' object is not iterable`).

However, it will already have placed some "stuff" into matplotlib's "queue of stuff to be drawn" (I have no idea what the terminology is here.)

Consequently, if you then fix your command and run it again:
```
wfdb.plot_wfdb(r, ecg_grids=[])
```
it'll display two windows (one with the half-finished result of the first command, one with the result of the second command.)  This is weird and confusing.

Assuming we can't magically anticipate all possible errors before invoking `plt.subplots`, can we "roll back" and cancel the effect of that function if an exception occurs?

PR #365, opened 2022-04-27
Close partially-completed plot windows after an error

bemoody:
If an error occurs while constructing a figure (e.g. because you passed invalid arguments to `plot_wfdb`), we should destroy the figure, rather than leaving a partially-completed figure in pyplot's internal queue where it will be unexpectedly displayed at some later time.  This should fix issue #347.

I think this'll do the right thing, but I've only tested it in the ordinary python interpreter; it should also be tested in jupyter.


cx1111 (approved):
Works in Jupyter. Thanks.

Issue #367, opened 2022-04-30
New interface for specifying different data sources for read/write


cx1111:
Looking at the current `rdrecord` for example, there are two parameters used to specify the location of the record:
1. `record_name` : str
2. `pn_dir` : str

The current package supports reading files locally and from the global database index URL, which defaults to PhysioNet, as specified in `download.py`.

There are several things that we should aim to support:
- Reading/writing from more types of data sources, such as S3, and GCS.
- Having more than one remote source configured at a time.

One proposal might be to have a new `DataSource` class, and a global config dictionary with key:value pairs of ds_name(str):ds(DataSource). ie.

```python
class DataSourceType(Enum):
    LOCAL = 1  # Not sure if this is necessary?
    HTTP = 2
    GCS = 3
    S3 = 4

class DataSource:
    ds_type : DataSourceType
    # Other type-specific params here
    
_physionet_ds = DataSource(ds_type=DataSourceType.HTTP, base_url="https://physionet.org/content/")

data_sources = { 'physionet' : _physionet_ds }

```

And the read/write functions could use these params:
1. `record_name`: str
2. `data_source`: str | DataSource - The key of the data source in the global data sources map, or a `DataSource` object.

This would be much more explicit. Thoughts?

PR #368, opened 2022-05-01
Add datasource module

cx1111:
Adding a new module to allow users to configure multiple explicit data sources.

Addresses #367

### DIFF
--- a/wfdb/io/datasource.py
+++ b/wfdb/io/datasource.py
@@ -1,0 +1,93 @@
+from enum import Enum, auto
+from pathlib import Path
+from typing import Dict
+from urllib.parse import urlparse
+
+
+class DataSourceType(Enum):
+    LOCAL = auto()
+    HTTP = auto()
+
+
+class DataSource:
+    def __init__(self, name: str, ds_type: DataSourceType, uri: str):
+        self.name = name
+        self.ds_type = ds_type
+        self.uri = uri
+
+    def __str__(self):
+        return f"{self.name} : {self.ds_type} : {self.uri}"
+
+    @property
+    def uri(self):
+        return self._uri
+
+    @uri.setter
+    def uri(self, value: str):
+        if self.ds_type == DataSourceType.LOCAL:
+            path = Path(value)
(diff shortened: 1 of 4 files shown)


cx1111:
These interfaces are very liable to change. We should finalize the initial version after adding at least one cloud data source.

Also, please give suggestions on associating credentialing information to these DataSources.

wfdb/io/datasource.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+_PHYSIONET_DATA_SOURCE = DataSource(
+    "physionet",
+    DataSourceType.HTTP,
+    "http://physionet.org/content/",
+)
+
+# Dict of configured data sources
+_data_sources: Dict[str, DataSource] = {"physionet": _PHYSIONET_DATA_SOURCE}
```
tompollard, line 49: Should _data_sources be all caps? Feels slightly odd to have this variable hanging around here.

wfdb/io/datasource.py
```diff
@@ -0,0 +1,93 @@
...
+                raise ValueError(
+                    "uri field for an HTTP DataSource must be a valid URL"
+                )
+        self._uri = value
+
+
+_PHYSIONET_DATA_SOURCE = DataSource(
+    "physionet",
+    DataSourceType.HTTP,
+    "http://physionet.org/content/",
```
tompollard, line 45: https?

tompollard (approved):
Looks good to me! Added a couple of minor comments, and it looks like there is a conflict that needs dealing with before we merge.

wfdb/io/datasource.py
cx1111: It's a global (module level) variable for storing the data sources. Not a constant.

wfdb/io/datasource.py
cx1111: Done.

cx1111:
> Why does this point to http://physionet.org/content/ and not https://physionet.org/files/?

Fixed

PR #370, opened 2022-05-02
Move conversion functions to their own modules and remove cyclic imports

cx1111:
The support for reading non-WFDB files is nice to have, but they should be recognized as extras, rather than being part of this library's core purpose.  I don't want to have to consider these other formats when developing/updating our core API. In particular, I'm trying to add a streaming API.

In addition, these functions were making the `record` module incredibly large. Therefore this PR moves a bunch of peripheral non-WFDB content to their own modules. It also renames some of functions to more appropriately. eg. `edf2mit` -> `read_edf` (because we're no longer using the "MIT" term and because the function didn't write files anyway)

Functionally, this also removes `rdrecord`'s ability to read EDF and WAV files. Users can use the individual functions `read_wav` and `read_edf` to get WFDB record objects instead.

Finally, a lot of content was moved to more appropriate modules to avoid cyclic dependencies.


wfdb/io/record.py
```diff
@@ -6,16 +6,11 @@
...
-import math
-import functools
-import struct
-import pdb
 
 from wfdb.io import _header
 from wfdb.io import _signal
 from wfdb.io import _url
 from wfdb.io import download
-from wfdb.io import annotation
```
cx1111, line 18: This looks like it used to be a circular import, and I don't know how it was working. See how `annotation.py` imports `record.py`. This was only here because of the `sigavg` function which I moved to the more appropriate `processing` subpackage.

wfdb/__init__.py
```diff
@@ -6,17 +6,10 @@
     rdsamp,
     wrsamp,
     dl_database,
-    edf2mit,
```
cx1111, line 9: Functions for the extra formats should not be part of the top level module.

cx1111:
Sorry I know this is really hard to review. I thought it'd be too time consuming to get multiple sequential reviews since there is so much stuff to move out of the way.

I've manually checked a reasonable amount by:
- Generating and looking through local Sphinx documentation to see if things have ended up in the right place.
- Using the `pydeps` tool to see import graphs. It seems to have some kinks in it though.
- Running the test suite locally and importing a few random things.

wfdb/io/record.py
cx1111: I don't quite understand how Python accepts/deals with cyclic deps... https://stackoverflow.com/questions/7336802/how-to-avoid-circular-imports-in-python

wfdb/io/convert/wav.py
```diff
@@ -0,0 +1,368 @@
...
+
+    sample_rate = struct.unpack("<I", wave_file.read(4))[0]
+    print("Sample rate: {}".format(sample_rate))
+    byte_rate = struct.unpack("<I", wave_file.read(4))[0]
+    print("Byte rate: {}".format(byte_rate))
+    block_align = struct.unpack("<H", wave_file.read(2))[0]
+    print("Block align: {}".format(block_align))
+    bits_per_sample = struct.unpack("<H", wave_file.read(2))[0]
+    print("Bits per sample: {}".format(bits_per_sample))
+    # I wish this were more precise but unfortunately some information
```
tompollard, line 293: Same comment as above. Should this be noted in the docstring?

Issue #336, opened 2021-11-15
Multi-frequency data with wfdb.wrsamp


bemoody:
If you have some multi-frequency data and are trying to save it as a WFDB record, currently the only way to do so is to create a Record object and call `Record.wrsamp`.  This is laborious and error-prone because of all the auxiliary fields that `Record.wrsamp` requires.

It'd be nice to be able to pass an `e_d_signal` or `e_p_signal` parameter to `wfdb.wrsamp` instead, e.g.:
```
wfdb.wrsamp(
    'example',
    fs=125,
    units=['mV', 'mmHg'],
    sig_name=['ECG', 'ABP'],
    samps_per_frame=[4, 1],
    e_p_signal=[
        numpy.array([0.282, 0.297, 0.304, 0.304,
                     0.311, 0.311, 0.333, 0.333]),
        numpy.array([70.093, 75.623])])
```

bemoody:
Related warts:

- `set_d_features(expanded=True, do_adc=True)` is broken (it stores the data incorrectly in `d_signal` rather than `e_d_signal`)

- `adc(expanded=True)` fails if `n_sig` is unset (it doesn't do that with `expanded=False`)

bemoody:
Fixed by pull #510.

Issue #333, opened 2021-10-28
wrsamp incorrectly saves samps_per_frame when expanded=False


bemoody:
When writing a record (header + signal file) using `Record.wrsamp`, the signal data may either be single-frequency (`d_signal`) or multi-frequency (`e_d_signal`).

If `expanded=True`, then `wrsamp` takes the data from `e_d_signal` (ignoring `d_signal`) and writes one or more multi-frequency signal files.  In this case the record must also have a `samps_per_frame` attribute, and this attribute is written to the header file.

If `expanded=False` (default), then `wrsamp` takes the data from `d_signal` (ignoring `e_d_signal`) and writes one or more single-frequency signal files.  In this case the `samps_per_frame` attribute is *still* written to the header file, which is completely incorrect.

Why would you have a Record that is single-frequency yet has `samps_per_frame` not equal to 1?  Because that's what `rdrecord` does by default when you *read* a multi-frequency record.

So:
```
>>> import wfdb
>>> r = wfdb.rdrecord('43700001', pn_dir='mimicdb/437', sampto=10, physical=0)
>>> r.wrsamp()
>>> wfdb.rdrecord('43700001')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/benjamin/wfdb-python/wfdb/io/record.py", line 3542, in rdrecord
    return_res=return_res)
  File "/home/benjamin/wfdb-python/wfdb/io/_signal.py", line 1010, in _rd_segment
    sampfrom, sampto, smooth_frames)[:, r_w_channel[fn]]
  File "/home/benjamin/wfdb-python/wfdb/io/_signal.py", line 1151, in _rd_dat_signals
    sig_data = _blocks_to_samples(sig_data, total_process_samples, fmt)
  File "/home/benjamin/wfdb-python/wfdb/io/_signal.py", line 1460, in _blocks_to_samples
    sig[0::2] = sig_data[0::3] + 256 * np.bitwise_and(sig_data[1::3], 0x0f)
ValueError: could not broadcast input array from shape (30) into shape (75)
```

(Arguably it's wrong for `rdrecord` to set `samps_per_frame` when `smooth_frames=True`, but changing that behavior would be hard to reconcile with `rdheader`, and would prevent implementing *automatic* frame-smoothing as I suggested in https://github.com/MIT-LCP/wfdb-python/pull/313#issuecomment-952033980)

PR #372, opened 2022-05-05
Read FLAC signal files

bemoody:
WFDB 10.7 defines new signal formats 508, 516, and 524.  Signal files in these formats are compressed using the FLAC algorithm.

Here, we add support for reading these formats using the soundfile package, which is a wrapper around libsndfile.  As such, you'll need to *have* libsndfile (and libFLAC.)  The soundfile wheel packages for Windows and MacOS include a copy of libsndfile that's statically linked with libFLAC.  On Linux or any other platform you will need to install libsndfile yourself, though there's a good chance you already have it.

Other implementation strategies are possible but this seems simplest.

The implementation in _signal.py is messier than I would like, because the _rd_dat_signals function combines low-level details (byte offsets, format-specific transformations) with high-level semantics (deskewing, length checking.)  Preferably the _skew_sig/_check_sig_dims stuff would be moved into _rd_segment instead.

Writing FLAC files is not implemented yet, and I think I'd like to try to address some of the other issues with wrsamp first (#333, #336).  Of course if somebody else wants to take a stab at it, feel free ;)


cx1111:
I see that this is still WIP. Can we all add type annotations to affected code from now on?

bemoody:
> Can we all add type annotations to affected code from now on?

It sounds like a good idea in principle, though it adds quite a bit of visual noise and we don't have an established style for doing so.  Could we take this discussion to a separate issue?

pyproject.toml
```diff
@@ -10,6 +10,7 @@ python = "^3.7"
 numpy = "^1.10.1"
 scipy = "^1.0.0"
 pandas = "^1.0.0"
+soundfile = "^0.10.0"
```
cx1111, line 13: I have trouble installing this with `poetry install`, which results in: `SolverProblemError  Because wfdb depends on soundfile (^0.10.0) which doesn't match any versions, version solving failed.`. Running `poetry add soundfile@^0.10.0` results in `SoundFile = "^0.10.0"` being added

https://pypi.org/project/SoundFile/

sample-data/flacformats.hea
```diff
@@ -0,0 +1,4 @@
+flacformats 3 200 499
```
cx1111, line 1: Do you have any multiplexed signal files for testing?

cx1111 (commented):
Looks clean. Please add a note in the README about installing WFDB and libsndfile.

wfdb/io/_signal.py
```diff
@@ -1718,6 +1738,159 @@ def _blocks_to_samples(sig_data, n_samp, fmt):
...
+        number `i*samps_per_frame[0]+k` of signal `j`.
+
+    Notes
+    -----
+    Converting the output array into "dat file order" here is inefficient,
+    but necessary to match the behavior of _rd_dat_file.  It would be
+    better to reorganize _rd_dat_signals to make the reshaping unnecessary.
+
+    """
+    import soundfile
```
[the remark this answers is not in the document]
cx1111: Oh, on M1 Mac when trying to `import soundfile`: `OSError: sndfile library not found`. Looks like a new release should support it: https://github.com/bastibe/python-soundfile/issues/310

Guess this conditional import is useful for now. Please write a note about M1 macs and reading flac files in the readme.

bemoody:
> I have trouble installing this with `poetry install`, which results
> in: `SolverProblemError  Because wfdb depends on soundfile (^0.10.0)
> which doesn't match any versions, version solving failed.`. Running
> `poetry add soundfile@^0.10.0` results in `SoundFile = "^0.10.0"`
> being added

Yeesh.  Can I just call that a bug in poetry?  Also, it looks like
soundfile 0.11.0b5 changed the name in setup.py from SoundFile to
soundfile.  I don't know if the name on PyPI follows the name in
setup.py or what.

bemoody:
> Oh, on M1 Mac when trying to `import soundfile`: `OSError: sndfile
> library not found`. Looks like a new release should support it:
> https://github.com/bastibe/python-soundfile/issues/310

Yeah, this is an annoyance for sure.  Did you try the (hot off the
presses) python-soundfile 0.11.0b5?

bemoody:
Bug: these formats need to be handled in _digi_nan.

bemoody:
I am completely lost when it comes to poetry.

If I check out 63c8039ffca8ce2b93489af0a06a64348d735f44, run `rm -rf v1; virtualenv -ppython3 v1; v1/bin/pip3 install poetry; v1/bin/poetry install`, I get the error you mentioned:
```
Installing dependencies from lock file
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.

  SolverProblemError

  Because wfdb depends on soundfile (^0.10.0) which doesn't match any versions, version solving failed.
```

If I do `v1/bin/poetry update soundfile; v1/bin/poetry install`, I get this nonsense:
```
Updating dependencies
Resolving dependencies... (0.2s)

Writing lock file

No dependencies to install or update
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: wfdb (4.0.0a0)
```
(it adds a bunch of junk to poetry.lock but it doesn't actually install any of the dependencies.)

The same happens if I do `git checkout pyproject.toml poetry.lock; git checkout 3b408f23f50889c7d1c6fee9329c76b739b9f667; rm -rf v1; virtualenv -ppython3 v1; v1/bin/pip3 install poetry; v1/bin/poetry install`, so clearly something was broken there to begin with.

If I rebase `flac` onto `main` (which has no `poetry.lock`) and do the same (`rm -rf v1; virtualenv -ppython3 v1; v1/bin/pip3 install poetry; v1/bin/poetry install`), it spins and spins.

In conclusion, I have no idea how to make poetry work, and I have no idea whether poetry will throw a fit if `SoundFile` changes its official spelling to `soundfile`.

If you have some idea how to make poetry work, please let me know the exact commands I ought to be running.

bemoody:
> If I rebase `flac` onto `main` (which has no `poetry.lock`) and do the same (`rm -rf v1; virtualenv -ppython3 v1; v1/bin/pip3 install poetry; v1/bin/poetry install`), it spins and spins.

Ooh, that command finally finished running after 14 minutes and it still didn't install anything.

cx1111:
I always remove `poetry.lock` before running `poetry install`, which may indicate other issues with me or  poetry... I've also experienced 20m of spinning in many occasions...

Anyway, I did manage to successfully install this config (removed big packages to save time) by running `poetry install` in a new virtual env without a `poetry.lock` file:
```
[tool.poetry.dependencies]
python = "^3.7"
numpy = "^1.10.1"
SoundFile = "^0.10.0"
matplotlib = "^3.3.4"
requests = "^2.8.1"
```

Running `pip install .` also worked.

I think if they change their name back to `soundfile` then we can update the name in our dependency file too.

wfdb/io/download.py
```diff
@@ -51,6 +51,70 @@ def set_db_index_url(db_index_url=PN_INDEX_URL):
...
+    mode="r",
+    *,
+    buffering=-1,
+    encoding=None,
+    errors=None,
+    newline=None,
+    check_access=False,
+):
+    """
+    Open a database file as a random-access file object.
```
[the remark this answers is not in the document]
cx1111: Also, perhaps this function should be in its own module, as `download.py` doesn't quite make sense. eg. Module `coreio.py` and function `open_file`.

cx1111:
> Oh, on M1 Mac when trying to `import soundfile`: `OSError: sndfile library not found`. Looks like a new release should support it: [bastibe/python-soundfile#310](https://github.com/bastibe/python-soundfile/issues/310)
> Yeah, this is an annoyance for sure. Did you try the (hot off the presses) python-soundfile 0.11.0b5?

I did not. Figured it wouldn't make a difference to this PR even if it works. I only tested your changes with the specified soundfile on amd64 Linux.

bemoody:
Thanks - it would be nice to check it works but not a top priority if that's difficult.

I haven't forgotten about the other issues you raised above and I'll try to address them tomorrow, but for now I've uploaded a short example of some real-world data you can play around with if you like.

bemoody:
Updated with the following changes:

- Add a note to README.md.

- Fix capitalization of SoundFile and add compatibility with 0.11.0 (soon to be released.)

- Rename `wfdb.io.download._open_db_file` to `wfdb.io._coreio._open_file`.

- Rewrite `_digi_nan` and `_digi_bounds` for clarity/completeness/efficiency and add FLAC formats.

Correctly handling invalid samples makes this example look better:
```
>>> import wfdb
>>> r = wfdb.rdrecord('sample-data/mixedsignals', sampto=1000)
>>> wfdb.plot_wfdb(r)
```
It still looks a little odd because the channels aren't synchronized (see pull #390).

bemoody:
By the way, I was poking around the other day and I saw that both `pulseaudio` and `pipewire-bin` packages on Debian have a dependency on `libsndfile1` - so you're very unlikely to find a desktop system that doesn't have libsndfile!  It's still worth mentioning in the README, though, because people will want to run the package on headless servers.

README.md
```diff
@@ -35,6 +35,8 @@ pip install wfdb
 poetry add wfdb
 ```
 
+On Linux systems, accessing *compressed* WFDB signal files requires installing `libsndfile`, by running `sudo apt-get install libsndfile1` or `sudo yum install libsndfile`.  Support for Apple M1 systems is a work in progess (see https://https://github.com/bastibe/python-soundfile/issues/310 and https://https://github.com/bastibe/python-soundfile/issues/325).
```
cx1111, line 38:
```suggestion
On Linux systems, accessing *compressed* WFDB signal files requires installing `libsndfile`, by running `sudo apt-get install libsndfile1` or `sudo yum install libsndfile`.  Support for Apple M1 systems is a work in progess (see <https://github.com/bastibe/python-soundfile/issues/310> and <https://github.com/bastibe/python-soundfile/issues/325>).
```

sample-data/mixedsignals.hea
```diff
@@ -0,0 +1,7 @@
+mixedsignals 6 62.4725/999.56 14400
```
cx1111, line 1: Looks like you haven't added a test case for these signals?

wfdb/io/_signal.py
```diff
@@ -1718,6 +1738,159 @@ def _blocks_to_samples(sig_data, n_samp, fmt):
...
+            elif sf.subtype == "PCM_16":
+                format_bits = 16
+                read_dtype = "int16"
+            elif sf.subtype == "PCM_24":
+                format_bits = 24
+                read_dtype = "int32"
+            else:
+                raise ValueError(f"unknown subtype in {fp.name} ({sf.subtype})")
+
+            max_bits = int(fmt) - 500
```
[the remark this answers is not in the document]
cx1111: Could you explain this logic?

bemoody:
> Looks like you haven't added a test case for these signals?

No, I didn't think of that.  Somebody want to add that? :P

> Could you explain this logic?

Well, one answer is just "trying to replicate the behavior of libFLAC
by unwrapping the layers that are added on top of it by libsndfile and
SoundFile."  But why?

libFLAC, like WFDB, deals with samples as integer values: an "8-bit"
FLAC file will be decoded as a series of integers between -128 and
127.  The "full range" is regarded as kind of an implementation detail
(fun fact, libFLAC doesn't enforce range limits on the input samples,
which can give weird results.)  Also, FLAC compression itself is
deeply tied to the quantization step.

SoundFile, on the other hand, conceptually regards samples as
fractions of the full range, and the quantization step is thought of
as an implementation detail.  If you ask for 32-bit integer data,
SoundFile assumes what you want is a value scaled to the range -2^31
to 2^31, regardless of the original storage format.  This is what you
generally want when working with audio.

So what we are doing is trying to persuade SoundFile to give us the
most efficient format we can get, while undoing any automatic
transformation it performs (which we can infer from the subtype
attribute).

As for the "max_bits" thing: the idea there is simply that you may
have a collection of FLAC files *not* produced by WFDB, and possibly
with unknown/varying bit depths, and you want to read them using WFDB.
So the format code is an upper bound and not a strict requirement.

tompollard:
@cx1111 are we good to merge this?

bemoody:
Thanks, Chen!  I fixed the typo in the README.  It would definitely be a good idea to add a test case for multiplexed/multi-frequency signals in the future but for right now I just want to merge the code as-is.

PR #375, opened 2022-05-17
Avoid using multiple processes for downloads

bemoody:
`multiprocessing` is messy; it's a *pretty nifty tool* but requires you to write your entire application with multiprocessing in mind.

For simply downloading files in parallel, there's little reason to use processes rather than threads.

This *should*, though I haven't tested it, solve the issues mentioned in pull #330 and probably also issue #306.  I can't guarantee that this will solve every problem, but it should be strictly better than what we have now.

### DIFF
--- a/wfdb/io/download.py
+++ b/wfdb/io/download.py
@@ -1,4 +1,4 @@
-import multiprocessing
+import multiprocessing.dummy
 import numpy as np
 import re
 import os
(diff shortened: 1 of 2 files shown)


tompollard:
@bemoody please could you fix the conflict?

tompollard (approved):
Thanks!

cx1111:
@lbugnon does this solve the windows downloading issue?

PR #379, opened 2022-05-27
Add __getitem__ to Record, allowing to copy a Record using slicing.

Ivorforce:
Supports convenient slicing across both time and space.

Functions like rdrecord support subselection of time and channels already. In many cases, that is sufficient.

However, a function like this might still be useful if:

* The file is not accessible from the current function; just a Record object is.
* A copy of a record should be made, with additional channels.

### Examples

```py
copy = record[:record.fs * 60 * 30, [0, 2, 5]]
copy.sig_name = ["x", "y", "z"]
copy.record_name = f"{record.record_name}-xyz"
copy.wrsamp()
```

### DIFF
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -2,6 +2,8 @@ import datetime
 import multiprocessing
 import posixpath
 import re
+import typing
+from copy import copy
 
 import numpy as np
 import os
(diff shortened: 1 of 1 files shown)


Ivorforce:
Closing for now, until better use-cases can be found.

PR #381, opened 2022-05-31
Add base_datetime property

bemoody:
`wfdb.rdrecord` currently sets an undocumented attribute `base_datetime`, but only if the record has a base date and a base time.

```
>>> r = wfdb.rdrecord('sample-data/n16')
>>> r.base_time
datetime.time(22, 34, 47)
>>> r.base_date
datetime.date(2006, 1, 1)
>>> r.base_datetime
datetime.datetime(2006, 1, 1, 22, 34, 47)

>>> r2 = wfdb.rdrecord('sample-data/3000003_0003')
>>> r2.base_time
datetime.time(19, 46, 25, 757000)
>>> r2.base_date
>>> r2.base_datetime
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Record' object has no attribute 'base_datetime'
```

It is frequently convenient to have this attribute available, so:

1. Always set `base_datetime`; set it to None if either `base_date` or `base_time` is None.

2. Define it as an aliased property so that setting `base_date` and `base_time` implicitly sets `base_datetime` and vice versa.

3. For consistency, also accept `base_datetime` as an argument to `wfdb.Record`, `wfdb.MultiRecord`, and `wfdb.wrsamp`.

Also, fix the documentation that incorrectly claimed `base_date` and `base_time` were strings.  They weren't.

### DIFF
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -208,6 +208,7 @@ class BaseRecord(object):
         sig_len=None,
         base_time=None,
         base_date=None,
+        base_datetime=None,
         comments=None,
         sig_name=None,
     ):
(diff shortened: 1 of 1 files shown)


bemoody:
Oh, incidentally it looks like this is required in order for pull #380 to work correctly - I didn't even notice that. :)

PR #387, opened 2022-06-07
Test cases for plot_wfdb

bemoody:
Add some test cases for wfdb.plot_wfdb.  Mostly we just check that it doesn't crash, but also check that it gives vaguely-reasonable X and Y plotting ranges.

plot.py coverage before: 571/978 lines

plot.py coverage after: 851/978 lines


tompollard (approved):
Looks good to me!

PR #388, opened 2022-06-07
[Draft]: Add new record

cx1111:


cx1111:
I considered having an individual `SignalHandler` or `Signal` class for each channel. But I think having one object represent all the channels makes more sense, especially regarding:
- Reading/writing multiplexed signal files. ie. A single signal file containing interleaved samples of multiple channels.

Issue #373, opened 2022-05-12
Plotting should use sharex=True by default


bemoody:
When plotting multiple signals (from the same record), they should be horizontally aligned with each other, even if the time extents are not the same.

An example we ran into the other day:
```
r = wfdb.rdrecord('sample-data/multi-segment/p000878/p000878-2137-10-26-16-57', channels=[0, 1], sampto=7000)
wfdb.plot_wfdb(r)
```
This plot doesn't make it obvious that the two signals don't start at the same time.  Compare:
```
wfdb.plot_items(r.p_signal, sharex=True)
```

In addition, setting `sharex=True` means that as you interactively scroll/zoom the plot, the time axis *stays* synchronized.

Therefore, I think `plot_wfdb` should, by default, invoke `plot_items` with `sharex=True`.  Maybe plot_wfdb should accept a `sharex` argument (or `share_time_axis`?) itself.

Note that as it stands, `sharex` is incompatible with `time_units='samples'` (or rather, it does completely the wrong thing for multi-frequency records.)  My suggestion would be to display frame numbers instead of sample numbers in this case, and/or to use seconds as the time unit by default.

PR #390, opened 2022-06-08
Align signals horizontally in plot_wfdb (if possible)

bemoody:
In `plot_wfdb`, add an argument `sharex` which controls whether X axes are shared between subplots.

(This means that multiple channels will always appear time-aligned with each other, even if one channel starts or ends with a block of NaNs.  It also means that if you use the pan/zoom buttons to navigate, the channels will *stay* aligned with each other.)

`sharex=True` is generally desirable behavior; however, if the time units are *sample numbers* and the record is multi-frequency, then it's not possible (AFAICT) with matplotlib to have the axes synchronized while displaying different units on different subplots.  So we disable sharex by default in that case.

Finally, `plot_wfdb` will use `time_units="seconds"` by default, which I think is useful because it's consistent (across databases) and familiar.

(original pull request description follows)

<s>A few changes to improve plotting of multiple channels and multiple frequencies:</s>

- <s>For `plot_items` and various internal functions, rename the `fs` argument to `frame_freq` so it is (perhaps) more clear what it means for multi-frequency data.  (I'm still not completely satisfied and think `plot_items` is harder to use than it needs to be.)</s>

- <s>Allow specifying `time_units="frames"` as an alternative to "samples", "seconds", etc.</s>

- <s>The big change: `plot_wfdb` uses `sharex=True`.  This means that multiple channels will always appear time-aligned with each other, even if one channel starts or ends with a block of NaNs.  It also means that if you use the pan/zoom buttons to navigate, the channels will *stay* aligned with each other.</s>

- <s>`plot_wfdb` will use `time_units="seconds"` by default, which I think is useful because it's consistent (across databases) and familiar.</s>

(fixes issue #373)

### DIFF
--- a/wfdb/plot/plot.py
+++ b/wfdb/plot/plot.py
@@ -863,6 +863,7 @@ def plot_wfdb(
     ecg_grids=[],
     figsize=None,
     return_fig=False,
+    sharex="auto",
 ):
     """
     Subplot individual channels of a WFDB record and/or annotation.
@@ -917,6 +918,12 @@ def plot_wfdb(
         'figsize' argument passed into matplotlib.pyplot's `figure` function.
     return_fig : bool, optional
         Whether the figure is to be returned as an output argument.
+    sharex : bool or 'auto', optional
+        Whether the X axis should be shared between all subplots.  If set
+        to True, then all signals will be aligned with each other.  If set
+        to False, then each subplot can be panned/zoomed independently.  If
+        set to 'auto' (default), then the X axis will be shared unless
+        record is multi-frequency and the time units are set to 'samples'.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


bemoody:
Whoops, fixed a typo.

wfdb/plot/plot.py
```diff
@@ -121,7 +121,7 @@ def plot_items(
     signal=None,
     ann_samp=None,
     ann_sym=None,
-    fs=None,
+    frame_freq=None,
```
cx1111, line 124: Demo 6 in `demo.ipynb` uses the old name.

cx1111, line 124: Demo 6 in `demo.ipynb` uses the old name.

wfdb/plot/plot.py
```diff
@@ -217,11 +216,11 @@ def plot_items(
     sampling_freq : number or sequence, optional
         The sampling frequency or frequencies of the signals.  If this is a
         list, it must have the same length as the number of channels.  If
-        unspecified, defaults to `fs`.
+        unspecified, defaults to `frame_freq`.
```
cx1111, line 219: Having these two parameters is rather confusing. Can you explain why both `sampling_freq` and `frame_freq` are both needed?

cx1111, line 219: Having these two parameters is rather confusing. Can you explain why both `sampling_freq` and `frame_freq` are both needed?

bemoody:
ACK, seems messier than I thought.

What if we forget about the frame_freq thing, forget about
sharex=True+units="samples", and instead just:

- use seconds by default in plot_wfdb
- use sharex=True in plot_wfdb *except* when units are set to samples

cx1111:
Yea, let's fix the `frame_freq` and `fs` issue separately. I propose:
1. Use seconds by default in `plot_wfdb`
2. Add a `sharex` parameter in `plot_wfdb` that can be:
  - "auto" (default): True, except when the signal is multi-frequency and units are set to samples.
  - True
  - False

cx1111 (approved):
Nice!

PR #393, opened 2022-06-15
Refactor header helper functions

cx1111:
Two main changes:
1. Remove the duplicate logic for parsing comment and non-comment lines from header text in `download.py` and `_header.py`
2. Fix the above logic to no longer try to find comment strings at the end of record/signal/segment lines. See: https://github.com/wfdb/wfdb-spec/pull/9#issuecomment-1121414020


tompollard (approved):
Looks good! So we are intentionally dropping support for inline comments? e.g. for the following data:

```
drive02.dat 16 1.0001/bpm 16 0 75 -19070 0 HR
drive02.dat 16 100 16 0 0 -9226 0 marker # INLINE COMMENT
drive02.dat 16x2 500 16 0 5804 -14191 0 RESP
# COMMENT ON OWN LINE
```
...previously we returned:

```
 'comments': ['INLINE COMMENT', 'COMMENT ON OWN LINE'],
 ```

...but now we return:

```
 'comments': ['COMMENT ON OWN LINE'],
```

It would be good to know how often inline comments appear in existing datasets (if at all), but fine with me if @bemoody is happy (which I assume he is based on https://github.com/wfdb/wfdb-spec/pull/9#issuecomment-1121414020).

PR #395, opened 2022-06-23
Add a missing import in _coreio.py

bemoody:
The posixpath module is used in _coreio.py but I mistakenly forgot to import it.  This breaks remote FLAC file access.


cx1111:
Also missing `config` object from `download` module.

cx1111 (approved):

Issue #394, opened 2022-06-22
Reading FLAC file fails when total samples > 2**24


bemoody:
Reading a FLAC signal file will fail when the total number of samples (length times number of channels) exceeds 2**24, *and* the number of channels is not a power of two.

The root cause appears to be a bug in libsndfile: the internal function flac_read_flac2s tries to break up the input into blocks of 0x1000000 total samples, which doesn't work so well if the number of channels isn't a power of two.

(It might also fail if the FLAC block size isn't a power of two.  Really, the logic in libsndfile is weird and I don't understand it.)

We should be able to work around this by reading a smaller block size at a time.

PR #397, opened 2022-06-24
Handle long-duration FLAC files

bemoody:
As described in issue #394, _rd_compressed_file fails when the number of samples to read exceeds 2**24 and the number of channels isn't a power of two.

This is a bug in libsndfile (https://github.com/libsndfile/libsndfile/issues/431), and although it should be fixed there and perhaps could be worked around in python-soundfile, it's most expedient to work around it here for now.


wfdb/io/_signal.py
```diff
@@ -1875,7 +1875,22 @@ def _rd_compressed_file(
...
+            # samples) due to a bug in libsndfile.  So read the data
+            # in chunks instead.
+            n_samp = end_samp - start_samp
+            sig_data = np.empty((n_samp, n_sig), dtype=read_dtype)
+            CHUNK_SIZE = 1024 * 1024
+            for chunk_start in range(0, n_samp, CHUNK_SIZE):
+                chunk_end = chunk_start + CHUNK_SIZE
+                chunk_data = sf.read(out=sig_data[chunk_start:chunk_end])
+                samples_read = chunk_data.shape[0]
+                if samples_read != CHUNK_SIZE:
```
cx1111, line 1891: What exactly is this block for? Is it just for the last chunk, or potentially earlier?

Why would the shape of `sig_data` need to be changed? Is its first dimension not guaranteed to be `n_samp`?

wfdb/io/_signal.py
```diff
@@ -1875,7 +1875,22 @@ def _rd_compressed_file(
...
+                    sig_data = sig_data[0 : chunk_start + samples_read]
```
cx1111, line 1892: Nit: Can drop the 0 here.

wfdb/io/_signal.py
```diff
@@ -1875,7 +1875,22 @@ def _rd_compressed_file(
             start_samp = start_frame * samps_per_frame[0]
             end_samp = end_frame * samps_per_frame[0]
             sf.seek(start_samp + sample_offset)
-            sig_data = sf.read(end_samp - start_samp, dtype=read_dtype)
+
+            # We could do this:
+            #  sig_data = sf.read(end_samp - start_samp, dtype=read_dtype)
+            # However, sf.read fails for huge blocks (over 2**24 total
+            # samples) due to a bug in libsndfile.  So read the data
+            # in chunks instead.
```
cx1111, line 1883: Please add links to any relevant GH issues here.

cx1111:
> The file might end up being shorter than expected. This can of course happen with binary files as well as FLAC. Currently, AFAIK, if the file is shorter than expected, then both _rd_compressed_file and _rd_dat_file will simply return a truncated array. _rd_dat_signals will then raise the dreaded "samples were not loaded correctly". We *could* raise an error in _rd_compressed_file instead, but my thought was that it would be better to handle it at a higher level. But if you would rather raise the error here, I can change that.

Got it, thanks. I think keeping the logic as is, being consistent with the existing implementations, makes sense.

cx1111 (approved):

PR #401, opened 2022-06-29
Split README content into separate DEVELOPING document

cx1111:
* Remove outdated readthedocs content

The reason for splitting the main README content is that the developing instructions shouldn't be propagated to other places such as pypi. ie. https://pypi.org/project/wfdb/4.0.0/


tompollard (approved):
Looks good to me!

Issue #418, opened 2022-08-31
Record.adc() truncates, doesn't round


bemoody:
```
>>> wfdb.wrsamp('aaa', fs=500, units=['mV'], sig_name=['X'], p_signal=numpy.array([[-1.001],[-0.999],[-0.001],[0.001],[0.999],[1.001]]), adc_gain=[1], baseline=[0], fmt=['16'])
>>> wfdb.rdsamp('aaa')[0]
array([[-1.],
       [ 0.],
       [ 0.],
       [ 0.],
       [ 0.],
       [ 1.]])
```
Clearly the result should be `[[-1],[-1],[0],[0],[1],[1]]`.

PR #419, opened 2022-09-01
Round sample values in physical-to-digital conversion

bemoody:
When converting physical (floating-point) values to digital (integer) values, we generally want to round to the nearest integer, not truncate towards zero.

(Translating floating-point data to integers is *more complicated than you'd think*, but we should try to avoid creating new sources of errors.)

One example:
```
>>> p = wfdb.rdrecord('sample-data/v102s', physical=True)
>>> d = wfdb.rdrecord('sample-data/v102s', physical=False)
>>> p.adc() == d.d_signal
array([[ True,  True,  True,  True],
       [ True,  True, False,  True],
       [ True,  True,  True,  True],
       ...,
       [ True,  True,  True,  True],
       [ True,  True,  True,  True],
       [ True,  True,  True,  True]])
```
(we would of course expect these arrays to be equal.)

This fixes issue #418.

*edit: I am not good at copy and paste*

### DIFF
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -971,6 +971,90 @@ class TestSignal(unittest.TestCase):
 
         assert record_2.__eq__(record)
 
+    def test_physical_conversion(self):
+        n_sig = 3
+        adc_gain = [1.0, 1234.567, 765.4321]
+        baseline = [10, 20, -30]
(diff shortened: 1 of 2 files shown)


tompollard (approved):
Looks good to me!

PR #420, opened 2022-09-02
FLAC output and assorted wrsamp improvements

bemoody:
This is another mixed bag of changes to `wfdb.wrsamp` and `Record.wrsamp`.

The main goal here is to allow writing compressed (FLAC) signal files.

Just as with reading signal files, this requires rearranging the input data (`d_signal` or `e_d_signal`) into the format required by the soundfile package.  Just as with reading signal files, this is not done as efficiently as it could be (in fact, none of the formats in `wr_dat_file` are implemented as efficiently as they could be.)

It's worth noting that `soundfile` doesn't provide a way to set the compression level or other parameters.  Setting the compression level, at least, should be easy to implement if somebody wanted to.  The `libsndfile` default appears to be level 5 (the same default as `libwfdb` and the `flac` command-line tool.)

When writing a record in FLAC format, it may be necessary to use multiple signal files (either because there are more than 8 channels or there are multiple sampling frequencies.)  `wfdb.wrsamp` doesn't provide a way to set the signal file names (which is likely a good thing; we don't need to over-complicate this API.)  On the other hand, `wfdb.wrsamp` *does* accept a *list* of file formats, even though it will fail if you try to use different formats for one signal file.  So here I have extended it to generate multiple signal files (`record_1.dat`, `record_2.dat`, etc.) if the formats require it.

In order to test reading and writing the sample record "flacformats", it was also necessary to fix another long-standing issue (`check_field` rejecting signal names containing spaces.)

Additionally, `Record.wrsamp` was broken for the case of expanded (multi-frequency) format and multiple signal files.  `wr_dat_file` will now sanity-check its arguments to try to catch such issues.

### DIFF
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -255,6 +255,23 @@ class TestRecord(unittest.TestCase):
         record_write = wfdb.rdrecord("flacformats", physical=False)
         assert record == record_write
 
+    def test_read_write_flac_multifrequency(self):
+        """
+        Format 516 with multiple signal files and variable samples per frame.
+        """
+        # Check that we can read a record and write it out again
+        record = wfdb.rdrecord(
+            "sample-data/mixedsignals",
+            physical=False,
+            smooth_frames=False,
+        )
+        record.wrsamp(expanded=True)
+
+        # Check that result matches the original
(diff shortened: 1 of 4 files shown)


briangow (approved):
@bemoody , this looks good to me, thanks!

PR #422, opened 2022-09-20
Store test-suite temporary files in temporary directories

bemoody:
When writing files as part of test cases, write them to a temporary
directory (created in setUpClass and destroyed in tearDownClass)
rather than writing them to the current working directory.

This allows running tests without modifying the working tree.  It
enables running tests in parallel using `pytest-xdist`, which allows
running the whole test suite in around 12 seconds rather than 35
seconds.

It's also just cleaner not to dump a lot of temporary files in the
working tree.

It does feel weird to be creating and deleting temporary directories
in *class* methods, but it seems to work.  (The alternative, using
setUp and tearDown, would mean creating a new directory for each
individual test function.)

### DIFF
--- a/tests/test_annotation.py
+++ b/tests/test_annotation.py
@@ -1,5 +1,6 @@
 import os
 import re
+import tempfile
 import unittest
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tompollard:
Looks great. [tempfile](https://docs.python.org/3/library/tempfile.html) is new to me. Good to know!

tompollard (approved):

PR #429, opened 2022-10-26
When reading edf, decode all strings using 'unicode_escape'.

Ivorforce:
The [edf specifications](https://www.edfplus.info/specs/edf.html) notes all strings being 'ascii'. 

By default, `decode()` uses the 'utf-8' encoding. In some cases, this can cause decoding issues:

```
File ~/Library/Caches/pypoetry/virtualenvs/workspace-lukas-kum-5zRav6QH-py3.9/lib/python3.9/site-packages/wfdb/io/convert/edf.py:262, in read_edf(record_name, pn_dir, header_only, verbose, rdedfann_flag)
    259 physical_dims = []
    260 for _ in range(n_sig):
    261     physical_dims.append(
--> 262         struct.unpack("<8s", edf_file.read(8))[0].decode().strip()
    263     )
    264 if verbose:
    265     print("Physical Dimensions: {}".format(physical_dims))

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb5 in position 0: invalid start byte
```

Arguably, the encoding `ascii` should be used. However, in my case, this didn't completely solve the problem:

```
File ~/Library/Caches/pypoetry/virtualenvs/workspace-lukas-kum-5zRav6QH-py3.9/lib/python3.9/site-packages/wfdb/io/convert/edf.py:262, in read_edf(record_name, pn_dir, header_only, verbose, rdedfann_flag)
    259 physical_dims = []
    260 for _ in range(n_sig):
    261     physical_dims.append(
--> 262         struct.unpack("<8s", edf_file.read(8))[0].decode("ascii").strip()
    263     )
    264 if verbose:
    265     print("Physical Dimensions: {}".format(physical_dims))

UnicodeDecodeError: 'ascii' codec can't decode byte 0xb5 in position 0: ordinal not in range(128)
```

Using the admittedly peculiar `unicode_escape` [encoding](https://docs.python.org/3/library/codecs.html#text-encodings) solved the problem for me. Using the verbose printing mode, the culprit can be seen (`µ`):

`Physical Dimensions: ['', 'mV', 'mV', 'mV', 'mV', 'mV', 'mV.s', 'µV', 'mV', 'mV.s', 'µV', '', 's', '', '', 'V']`

The file in question was produced by [LabView](https://www.ni.com/de-de/shop/labview.html). Unfortunately, I cannot share it. If need be, it might be possible to reproduce the error otherwise.

### DIFF
--- a/wfdb/io/convert/edf.py
+++ b/wfdb/io/convert/edf.py
@@ -20,6 +20,7 @@ def read_edf(
     header_only=False,
     verbose=False,
     rdedfann_flag=False,
+    encoding="iso8859-1",
 ):
     """
     Read a EDF format file into a WFDB Record.
@@ -61,6 +62,9 @@ def read_edf(
         is being called by the user and the file has annotations, then warn
         them that the EDF file has annotations and that they should use
         `rdedfann` instead.
+    encoding : str, optional
+        The encoding to use for strings in the header. Although the edf
+        specification requires ascii strings, some files do not adhere to it.
 
     Returns
     -------
@@ -139,7 +143,7 @@ def read_edf(
     edf_file = open(record_name, mode="rb")
 
     # Version of this data format (8 bytes)
-    version = struct.unpack("<8s", edf_file.read(8))[0].decode()
+    version = struct.unpack("<8s", edf_file.read(8))[0].decode(encoding)
 
     # Check to see that the input is an EDF file. (This check will detect
     # most but not all other types of files.)
@@ -152,7 +156,7 @@ def read_edf(
             print("EDF version number: {}".format(version.strip()))
 
     # Local patient identification (80 bytes)
-    patient_id = struct.unpack("<80s", edf_file.read(80))[0].decode()
+    patient_id = struct.unpack("<80s", edf_file.read(80))[0].decode(encoding)
     if verbose:
         print("Patient ID: {}".format(patient_id))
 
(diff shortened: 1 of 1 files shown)


bemoody:
Is `decode("unicode_escape")` somehow different from `decode("iso8859-1")`?  It looks to me like it does the same thing.

I think if you find a random EDF file, it's more likely to be encoded in ISO-8859-1 (or Windows-1252) than UTF-8, but it'd be nice for the wfdb package to be "locale-neutral".

Could we add 'encoding' and maybe 'errors' arguments to read_edf, to make this explicit?  Maybe with ISO-8859-1 or Windows-1252 as default.

Ivorforce:
Yep, seems like `iso8859-1` works too. I admit i blindly copied the encoding from some Stackoverflow issue, eager to go forward, but reading up on it being python specific, it really doesn't make sense to use it.

I like your suggestion to make encoding a parameter. I'm updating the PR accordingly.

bemoody:
It looks good to me.  Thanks!  I don't know why the test is failing :/

bemoody:
Oh wait, it looks like you need to add the 'encoding' parameter to rdedfann() as well.

Ivorforce:
Right, makes sense. Fixed the issue.

bemoody:
Looks great!

PR #433, opened 2022-11-18
Correct example code for rdheader

slishak:
Noticed that the example code for `wfdb.rdheader` seemed to be out of date - updated it to match the actual function signature.


bemoody:
Thanks!

PR #435, opened 2022-11-21
Changes for CHB-MIT Scalp EEG Database

FelipeMLopez:
Hi!
I have encountered some problems reading the recordings from this dataset and have created these work-arounds.
In the comments I have described the problem in each case.
Maybe it's too specific for that dataset, but I leave it here in case it helps someone.
Regards.

### DIFF
--- a/wfdb/io/convert/edf.py
+++ b/wfdb/io/convert/edf.py
@@ -290,21 +290,34 @@ def read_edf(
 
     # Digital minimum (e.g., -2048) (8 bytes each)
     digital_min = np.array([])
-    for _ in range(n_sig):
+    for ii in range(n_sig):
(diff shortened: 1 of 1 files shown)


FelipeMLopez:
You can reproduce the log2 problem by trying:
```python
record = read_edf('chb16_01.edf', pn_dir=f'chbmit/chb16', header_only=False, verbose=True, rdedfann_flag=False)
```
And the reshape problem with:
```python
record = read_edf('chb17b_69.edf', pn_dir=f'chbmit/chb17', header_only=False, verbose=False, rdedfann_flag=False)
```

tompollard:
@FelipeMLopez this sounds like an issue with the [CHB-MIT Scalp EEG Database](https://doi.org/10.13026/C2K01R), rather than the WFDB package itself? If so, I think we should probably be asking the authors of the database to fix the data, rather than adding workarounds to the WFDB software.

PR #437, opened 2022-12-02
Release version 4.1.0

bemoody:
Version number and documentation updates for 4.1.0.

I've added the file changes.rst.  I think it is important to keep track of, and document, user-visible changes, and I think this document should be kept up to date *as new changes are committed*, not just when assembling a final release.

So my first thought was to ask people to update this file every time they make a pull request.  But that would create merge conflicts (& even if the person merging the pull can resolve those conflicts, github can't do so itself and therefore this would break workflow checks.)  And I don't want to start encouraging people to merge or cherry-pick changes from the main branch into their feature branch (rebasing would be fine, but for some reason a lot of people seem to be allergic to rebasing.)  So... I don't have a good solution here, and would welcome suggestions. :)


tompollard (approved):
Thanks @bemoody looks good to me.

> So my first thought was to ask people to update this file every time they make a pull request. But that would create merge conflicts (& even if the person merging the pull can resolve those conflicts, github can't do so itself and therefore this would break workflow checks.) 

I can't think of a clever way of managing this, other than trying to remind people to update the changelog immediately before merging. One thing that might help is adding some kind of reminder to a pull request template? 

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

PR #440, opened 2023-01-25
Fix code style

bemoody:
We decided a while ago to require code in this project to be formatted using `black`.

Fix the style of `edf.py` accordingly.  Require correct formatting for future pull requests.


tompollard (approved):
Looks good to me!

PR #447, opened 2023-04-14
Allow pandas >= 2

### LINKED ISSUE #446: Support for Pandas 2.0?
Hello, currently, Pandas is pinned to version 1.x:

https://github.com/MIT-LCP/wfdb-python/blob/4d4233fba53a96cdb6d2064c4e4ba0131c78944f/pyproject.toml#L16

Meanwhile, Pandas 2.0 has been released, and trying to install it alongside `wfdb` logically leads to a dependency conflict.

I wanted to ask if `wfdb` is supposed to not work with Pandas 2.0 now or if the version pinning is unnecessarily restrictive and could be relaxed to include Pandas 2.0 as well?

Thanks,
Richard

cbrnr:
Fixes #446. I've set the lower bound rather arbitrarily to 1.3.0 (which is about 2 years old). Let me know if that's OK or if you prefer a different lower bound.


tompollard:
Thanks @cbrnr, looks good to me!

PR #448, opened 2023-04-14
Remove version caps

cbrnr:
Currently, all versions are capped, because the caret `^` operator also introduces an upper bound. In this PR, I replaced `^` with `>=`, which continues to use the same lower bounds, but removes any upper bound.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,18 +10,18 @@ documentation = "https://wfdb.readthedocs.io/"
 license = "MIT"
 
 [tool.poetry.dependencies]
-python = "^3.7"
-numpy = "^1.10.1"
-scipy = "^1.0.0"
+python = ">=3.7"
+numpy = ">=1.10.1"
+scipy = ">=1.0.0"
 pandas = ">=1.3.0"
-SoundFile = ">=0.10.0, <0.12.0"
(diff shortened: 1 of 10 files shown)


tompollard:
Thanks @cbrnr, please could you rebase? Not immediately clear why the tests are failing. Looks like they are failing on a style issue in a chunk of unrelated code?

cbrnr:
Done. It's because the previous pandas-related change interferred with these modifications.

tompollard:
@bemoody Please could you add your thoughts on this PR? I see that SoundFile is explicitly pinning an upper bound (`SoundFile = ">=0.10.0, <0.12.0"`), which I assume is intentional? Does this need to stay?

@cbrnr Presumably the change to Black is causing the tests to fail. We need to look into what's happening there!

cbrnr:
I've fixed all Black issues, these were related to empty lines that shouldn't be there. Seems like they updated their style guide, because I noticed the same in one of my projects recently.

Let me know about the `SoundFile` version. If you really need to keep `<0.12.0`, we should add a test.

tompollard (approved):
Thanks @cbrnr, looks good to me! @bemoody?

PR #450, opened 2023-05-11
Handle more than 8 compressed signals in wrsamp

bemoody:
@alistairewj pointed out that there doesn't seem to be an easy way to write a compressed record with more than 8 channels.

It's supposed to be possible to do this:
```
import numpy, wfdb

wfdb.wrsamp('aaa', fs=1000, units=['mV'] * 12,
            sig_name=[str(i) for i in range(12)],
            p_signal=numpy.random.rand(100, 12),
            fmt=['516'] * 12)
```
but the logic I introduced in pull #420 is buggy and currently raises an error.


tompollard (approved):
thanks benjamin

PR #451, opened 2023-05-11
Add a test to check if wrsamp can write more than 8 channels to format 516

alistairewj:
We should be able to read and write data with more than 8 channels to formats 508, 516, and 524. This test verifies that the digital signal generated when writing to format 516 is identical to the original signal loaded in from a sample record in format 16.

Should catch bugs like those fixed in #450

### DIFF
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -286,6 +286,33 @@ class TestRecord(unittest.TestCase):
         )
         assert record == record_write
 
+    def test_read_write_flac_many_channels(self):
+        """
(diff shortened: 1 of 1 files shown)


tompollard:
@alistairewj more tests are always good! looks like the tests failed with a formatting issue. please could you check format and update the PR?

alistairewj:
Ah OK I had autopep8 running and not black. Updated now, fingers crossed!

tompollard (approved):
Looks good to me, thanks!

Issue #453, opened 2023-05-16
error when running processing.ann2rr


BarAbba39:
228     out_interval = time_interval / (60 * 60)
    229 else:
--> 230     out_interval = np.around(time_interval * ann.fs).astype(**np.int)**
    232 if as_array:
    233     return out_interval

File ---/venv/lib/python3.10/site-packages/numpy/__init__.py:284, in __getattr__(attr)
    281     from .testing import Tester
    282     return Tester
--> 284 raise AttributeError("module {!r} has no attribute "
    285                      "{!r}".format(__name__, attr))

error: np has no attribute np.int

np.int was deprecated in 2021

tompollard:
This issue should be fixed in Version 4.1.2, now available on PyPI. 

Please update your installation with:

```
pip install wfdb --upgrade
```

PR #454, opened 2023-05-25
Use int64 instead of int for ann2rr

bemoody:
This fixes issue #453 by using 64-bit integers on all platforms.  (64-bit integers are way overkill for storing RR intervals, but I think it's wise to keep it simple and use 64-bit integers for all types of sample intervals.)

This is a pretty simple fix, but looking at the code I also notice there are a couple of significant issues with this function:

- Unlike the ann2rr program, this function *only* considers the timestamps of the input annotations - it doesn't check whether they're QRS annotations or some other type of annotations (e.g. NOTE or RHYTHM).

- *Like* the ann2rr program, this function includes the interval between the first annotation and the start of the record.

As a result, when you look at Lucas's example in the docstring:
```
    >>> wfdb.ann2rr('sample-data/100', 'atr', as_array=False)
    >>> 18
    >>> 59
    >>> ...
    >>> 250
    >>> 257
```
the first two "RR intervals" listed here are bogus and misleading: the first is the interval between the start of the record and the first RHYTHM annotation, and the second is the interval between the RHYTHM annotation and the first NORMAL annotation.

All that is by way of saying, I'd like to add a test case for this function but I think it's kind of broken to begin with and probably its behavior should be better nailed down first.

### DIFF
--- a/wfdb/processing/hr.py
+++ b/wfdb/processing/hr.py
@@ -227,7 +227,7 @@ def ann2rr(
     elif format == "h":
         out_interval = time_interval / (60 * 60)
     else:
-        out_interval = np.around(time_interval * ann.fs).astype(np.int)
+        out_interval = np.around(time_interval * ann.fs).astype(np.int64)
 
     if as_array:
         return out_interval


tompollard:
Looks good to me, thanks Benjamin.

tompollard (approved):

PR #460, opened 2023-06-27
bug-fix: Pandas set indexing error

ajadczaksunriselabs:
Current version of Pandas will throw the following error when indexing using a set.

TypeError: Passing a set as an indexer is not supported. Use a list instead.


ajadczaksunriselabs:
I think this is good to go now.  Pandas started enforcing this behavior for the 2.0 release.
https://github.com/pandas-dev/pandas/pull/49511

PR #465, opened 2023-09-21
fix for MIT-LCP/wfdb-python/issues/452

tecamenz:
Fix for MIT-LCP/wfdb-python/issues/452

Using os.sep on Windows fails here because we are specifically searching for "/" as a separator


bemoody:
Looks good.  Thanks!

bemoody (approved):
oops i was supposed to review first

PR #475, opened 2023-12-17
handling local headers of mit bih database

Cheppy:
this would save time when working locally with mit-bih database

### DIFF
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -1835,14 +1835,22 @@ def rdheader(record_name, pn_dir=None, rd_segments=False):
 
     # Read the local or remote header file.
     file_name = f"{base_record_name}.hea"
+    file_name_mit_bih = f"{base_record_name}.dat.hea"
     if pn_dir is None:
-        with open(
-            os.path.join(dir_name, file_name),
-            "r",
-            encoding="ascii",
-            errors="ignore",
-        ) as f:
-            header_content = f.read()
+        file_from_mit_bih = os.path.join(dir_name, file_name_mit_bih)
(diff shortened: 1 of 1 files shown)


bemoody:
Please explain what you're trying to do.  The change you're suggesting does not make any sense to me.

Cheppy:
> Please explain what you're trying to do. The change you're suggesting does not make any sense to me.

Greetings, what I am trying to do:
There is open ECG database that is most commonly used to train datasets, or to check accuarcy of "own methods" for peak detection:
https://www.physionet.org/content/mitdb/1.0.0/
 The wfdb package currently only recognizes header files with the .hea extension, but the MIT-BIH database uses the .dat.hea format. To avoid renaming of headers or writing a script to rename them, I'm proposing a way to handle the case with .dat.hea files directly within wfdb.  This will help saving time and effort for those working with that database. and i think it would be nice idea, because MIT-BIH is commonly used in many scientific papers.
Maybe my way of handling that is too simple, but i hope you got the idea. What you think, should this case be added to wfdb?

bemoody:
Header files just have a `.hea` extension - `100.dat.hea` is not valid.  For example, the header file for record `100` is [`100.hea`](https://www.physionet.org/content/mitdb/1.0.0/100.hea).

You can load this record using `wfdb.rdrecord(record_name="100", pn_dir="mitdb/1.0.0")`.

Or, if you have downloaded the files `100.hea` and `100.dat`, you can use `wfdb.rdrecord("100")`.

I don't know where you found ".dat.hea" files; there aren't any on PhysioNet.

Issue #480, opened 2024-03-28
Record.adc: RuntimeWarning: invalid value encountered in cast


bemoody:
The `adc` function reports a RuntimeWarning (at least in some cases) if the input array contains NaN values.

I'm currently looking at this case (expanded=True, inplace=False), though it looks like the other cases are also broken in various similar ways:
```
                d_signal = []
                for ch in range(self.n_sig):
                    # NAN locations for the channel
                    ch_nanlocs = np.isnan(self.e_p_signal[ch])
                    ch_d_signal = self.e_p_signal[ch].copy()
                    np.multiply(ch_d_signal, self.adc_gain[ch], ch_d_signal)
                    np.add(ch_d_signal, self.baseline[ch], ch_d_signal)
                    np.round(ch_d_signal, 0, ch_d_signal)
                    ch_d_signal = ch_d_signal.astype(intdtype, copy=False)
                    ch_d_signal[ch_nanlocs] = d_nans[ch]
                    d_signal.append(ch_d_signal)
```
The warning occurs here:
```
                    ch_d_signal = ch_d_signal.astype(intdtype, copy=False)
```
numpy is complaining, I assume, because you can't cast a NaN to an integer.

(Like many things to do with numpy and with this package), the above looks rather inefficient.  Perhaps `numpy.nan_to_num` could be used instead.

bemoody:
> Perhaps numpy.nan_to_num could be used instead.

Never mind: `nan_to_num` is not any more intelligently implemented (might even be worse) than what wfdb-python is already doing.

bemoody:
To reproduce the issue:
```
import wfdb, numpy
r = wfdb.Record(fs=500, n_sig=1, units=['mV'], sig_name=['ECG'], p_signal=numpy.array([[0], [1], [numpy.nan]]), fmt=['16'], adc_gain=[1], baseline=[0])
r.adc()
```
or
```
import wfdb, numpy
r = wfdb.Record(fs=500, n_sig=1, units=['mV'], sig_name=['ECG'], e_p_signal=[numpy.array([0, 1, numpy.nan])], fmt=['16'], adc_gain=[1], baseline=[0])
r.adc(expanded=True)
```

This is fixed with pull #481.

PR #481, opened 2024-04-18
Fix NaN handling in Record.adc, and other fixes

bemoody:
Fix several bugs in `Record.adc`:

1. Previously, the function would try to convert all samples to integers and then, for any samples that were NaN, replace the corresponding elements with the appropriate sentinel value.  Even though this was probably safe in most cases, casting NaN to an integer is implementation-defined behavior, and raises a warning by default (issue #480).

2. NaN just plain wasn't handled for the `inplace=True, expanded=False` case.  (Currently, we don't use `inplace=True` anywhere internally; although it saves a bit of memory, it's destructive and so it's probably wise for high-level functions like `wrsamp` to avoid it.)

3. The `expanded=True` case relied on `self.n_sig` (in contrast to `expanded=False`, which operates based on the dimensions of `p_signal`.)  This meant it would fail if the caller didn't explicitly set `n_sig`, which was an annoying inconsistency.

Also, tidy up duplicated code and make things a little more efficient.

A side note: I don't think the `inplace=True` mode is particularly great to have.  It conflates two things (modifying the Record object attributes, which many applications want; and modifying the array contents, which you may think you want until you realize it subtly breaks something.)  It does save some memory, but not as much as you'd hope.  (That `copy=False` is pretty much a lie.)  And of course I don't like functions whose return type is dependent on their arguments.  So I would definitely put `inplace` on the chopping block for 5.0.0.  Still, I think the updated code here isn't too terribly ugly.

This set of changes is the first step to making `wfdb.wrsamp` work for multi-frequency (issue #336).  Next is to fix `Record.calc_adc_params`, then `Record.set_d_features`.

### DIFF
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -1053,19 +1053,20 @@ class TestSignal(unittest.TestCase):
         adc_gain = [1.0, 1234.567, 765.4321]
         baseline = [10, 20, -30]
         d_signal = np.repeat(np.arange(-100, 100), 3).reshape(-1, 3)
+        d_signal[5:10, :] = [-32768, -2048, -128]
         e_d_signal = list(d_signal.transpose())
-        fmt = ["16", "16", "16"]
+        fmt = ["16", "212", "80"]
 
         # Test adding or subtracting a small offset (0.01 ADU) to check
         # that we correctly round to the nearest integer
         for offset in (0, -0.01, 0.01):
             p_signal = (d_signal + offset - baseline) / adc_gain
+            p_signal[5:10, :] = np.nan
             e_p_signal = list(p_signal.transpose())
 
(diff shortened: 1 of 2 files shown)


tompollard:
@bemoody this looks good to me. were the tests intentionally stopped?

tompollard:
Hmm, it looks like maybe it is failing on format checks (https://github.com/MIT-LCP/wfdb-python/actions/runs/8744000168/job/24038235919) and then the remaining tests are being cancelled.

bemoody:
Yeah, I think it should work, it's failing because the latest version of black requires formatting changes (not related to this pull AFAIK.)

tompollard (approved):

PR #489, opened 2024-06-06
Fix: Indent code to ensure 'j' is within for-loop in GQRS algorithm

Biboulder:
The variable j was not properly scoped within the for-loop in the sm method, leading to an UnboundLocalError when j was accessed outside the loop.


tompollard:
@Biboulder Thanks for picking this up! Please could you rebase the pull request on the main branch, and if possible add a simple test that covers the `UnboundLocalError`?

tompollard:
I rebased on main and opened a new PR: https://github.com/MIT-LCP/wfdb-python/pull/499

PR #491, opened 2024-06-27
Add the ability to pass sig and hea streams to rdrecord and rdheader

Ivorforce:
… in case the file is not read from disk.

This is needed if, for example, files are streamed using a file upload dialog versus read from the file system. Personally, I ran into this limitation using Dash's Upload widget.

There are some basic ToDos:
- [ ] Documentation
- [ ] Unit Tests (?)

However, with this PR the code is starting to get a bit messy. In my opinion, some choices should be reconsidered before merge by code design. I propose the following changes (happy to oblige within the PR if accepted):
- Deprecate no_file field, instead use a `None` test against sig_data
- Since all paths end up with a byte stream anyway, lift the stream generation out to the public methods. private methods should know only streams, not file systems.

Let me know what you think.


Ivorforce:
Update: I have added the same cabability to `rdann`. 

`rdann` seems to have an additional problem right now of simply not working. A simple `.astype(np.int32)` seems to fix the issue, but may not be what's actually needed.

Ivorforce:
Update 2: I have opened #498. That PR is a more comprehensive and streamlined rewrite, and thus supersedes it, in my opinion.

Issue #490, opened 2024-06-26
csv_to_wfdb writes to current dir


jshaffer94247:
csv_to_wfdb currently strips the path from the input <basename>.csv, then writes the output to <basename>.dat and <basename>.hea

Request is for
a) full path to be preserved for the output file (not backward compatible)
b) accept an optional parameter to specify the output folder (preserves backward compatibility)

tompollard:
@jshaffer94247 not ideal, it should definitely be possible to specify the output directory! 

Does adding an `output_dir` argument sound okay, or do you have a cleaner suggestion?

```
def csv_to_wfdb(
    ...
    delimiter=",",
    verbose=False,
    output_dir=None,
):
```

jshaffer94247:
I do not have a cleaner suggestion; that's what I'd do.

tompollard:
`write_dir` will be available in the next release. If you'd like to use it now, you should be able to pip install from this repo with: 

```
pip install git+https://github.com/MIT-LCP/wfdb-python.git@6a0de803b4d396fa22da583dfa0d19250d347217
```

PR #492, opened 2024-07-01
Add write_dir argument to csv_to_wfdb. Fixes #67.

### LINKED ISSUE #67: Memory optimization
Hello, I am facing some memory problems when reading big databases (which I would like to store completly in the memory). I noticed that the wfdb functions generally return float64 numpy arrays. However, if I understood correctly from the link below, the signals are coded on maximum 32 bits.
https://www.physionet.org/physiotools/wag/signal-5.htm

Here are my questions / remarks:
- does this mean it is safe to convert all signals to 32bits ? 
- if this is safe, would it be possible to make your functions return float32 arrays ? Or even better to return the minimum size format depending on the header format field ?

G.

tompollard:
As discussed in https://github.com/MIT-LCP/wfdb-python/issues/490, https://github.com/MIT-LCP/wfdb-python/blob/34b989e08435c1a82d31bdd2800c4c14147e3e93/wfdb/io/convert/csv.py#L10 currently "strips the path from the input .csv, then writes the output to .dat and .hea".

It's inconvenient not to be able to specify the output directory. This pull request adds a new `output_dir` argument to the `csv_to_wfdb` function. By default `output_dir` is set to None, which will maintain backwards compatibility. Setting `output_dir` to a directory will mean that output files are saved to this directory.

I have set this to a WIP, because I haven't tested the new behaviour (other than running `pytest`). @jshaffer94247, if you have an opportunity to test the fix, I'd appreciate your feedback.

### DIFF
--- a/tests/io/test_convert.py
+++ b/tests/io/test_convert.py
@@ -1,14 +1,22 @@
+import os
+import shutil
+import unittest
+
 import numpy as np
 
 from wfdb.io.record import rdrecord
 from wfdb.io.convert.edf import read_edf
+from wfdb.io.convert.csv import csv_to_wfdb
 
 
-class TestConvert:
+class TestEdfToWfdb:
+    """
+    Tests for the io.convert.edf module.
+    """
+
     def test_edf_uniform(self):
         """
         EDF format conversion to MIT for uniform sample rates.
-
         """
         # Uniform sample rates
         record_MIT = rdrecord("sample-data/n16").__dict__
@@ -60,7 +68,6 @@ class TestConvert:
     def test_edf_non_uniform(self):
         """
         EDF format conversion to MIT for non-uniform sample rates.
-
         """
         # Non-uniform sample rates
         record_MIT = rdrecord("sample-data/wave_4").__dict__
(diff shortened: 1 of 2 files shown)


tompollard:
Tests are failing on an unrelated issue, presumably related to an update to Numpy (e.g. this? https://numpy.org/devdocs/release/1.24.0-notes.html#)

bemoody:
It should be `write_dir` to be consistent with `wrsamp`, `wrann`, etc.

The existing logic is pretty broken, though!  `split(os.sep)` is decidedly wrong for Windows, but also `replace(".csv", "")` is screwy.  You want to take the basename and then remove *anything* after a dot.

bemoody:
No test cases - please add a test case so we can see that this function works.

tompollard:
@bemoody this is now ready to review.

bemoody:
`pandas/core/base.py` says "tolist is not deprecated" (`to_list` is an alias for `tolist`), so can we simply use `tolist`?

tompollard:
> `pandas/core/base.py` says "tolist is not deprecated" (`to_list` is an alias for `tolist`), so can we simply use `tolist`?

I considered this but decided to use `to_list()` because it sounds like it is what the developers would like us to use (I think for consistency with other pandas methods). I think dropping the try/except is fine though...now done.

bemoody (approved):
Thanks - looks good!

tompollard:
Thanks Benjamin

Issue #493, opened 2024-07-03
OverflowError: Python integer 256 out of bounds for uint8


cve2022:
I am breaking the demos in the notebook in individual to get acquainted with the package to learn how it works and adapt it to my needs.

Running demo4 as a separate script with python I get this error traceback:

```
D:\Users\xxx\Work\__WFDB\wfdb-python>python demo4.py
Traceback (most recent call last):
  File "D:\Users\xxx\Work\__WFDB\wfdb-python\demo4.py", line 12, in <module>
    annotation = wfdb.rdann('sample-data/100', 'atr', sampfrom=100000, sampto=110000)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\xxx\Work\__WFDB\wfdb-python\wfdb\io\annotation.py", line 1953, in rdann
    (sample, label_store, subtype, chan, num, aux_note) = proc_ann_bytes(
                                                          ^^^^^^^^^^^^^^^
  File "D:\Users\xxx\Work\__WFDB\wfdb-python\wfdb\io\annotation.py", line 2154, in proc_ann_bytes
    sample_diff, current_label_store, bpi = proc_core_fields(filebytes, bpi)
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\xxx\Work\__WFDB\wfdb-python\wfdb\io\annotation.py", line 2240, in proc_core_fields
    sample_diff += int(filebytes[bpi, 0] + 256 * (filebytes[bpi, 1] & 3))
                                           ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
OverflowError: Python integer 256 out of bounds for uint8
```

The same happens with Linux (Ubuntu 22.04).

After changing the highlighted line it works like the notebook.
```
$ git diff
diff --git a/wfdb/io/annotation.py b/wfdb/io/annotation.py
index b398fa0..4d451b1 100644
--- a/wfdb/io/annotation.py
+++ b/wfdb/io/annotation.py
@@ -2237,7 +2237,7 @@ def proc_core_fields(filebytes, bpi):

     # Not a skip - it is the actual sample number + annotation type store value
     label_store = filebytes[bpi, 1] >> 2
-    sample_diff += int(filebytes[bpi, 0] + 256 * (filebytes[bpi, 1] & 3))
[...]

PR #494, opened 2024-07-03
Set upper bound on Numpy version (numpy = ">=1.10.1,<2.0.0"). Ref #493.

tompollard:
As discussed in #493, numpy v2.0.0 introduces a breaking change for WFDB: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion. 

This pull request sets an upper bound on the numpy version as a temporary fix (`numpy = ">=1.10.1,<2.0.0"`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ license = "MIT"
 
 [tool.poetry.dependencies]
 python = ">=3.7"
-numpy = ">=1.10.1"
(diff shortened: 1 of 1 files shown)


bemoody (approved):
Seems like the best temporary fix.

tompollard:
Thanks Benjamin

PR #497, opened 2024-07-03
Bump version to v4.2.0 and update notes on creating new releases

tompollard:
This is a minor change that:

- Bumps the version number to 4.1.3
- Updates the notes on creating new releases

After this change we will need to:

- Update the [changelog](https://github.com/MIT-LCP/wfdb-python/blob/main/docs/changes.rst)
- Update the [documentation](https://github.com/MIT-LCP/wfdb-python/tree/main/docs) if necessary
- Create a new release at: https://github.com/MIT-LCP/wfdb-python/releases/new


bemoody:
I'm okay with cutting a new release here - I think addressing the FLAC bugs is unfortunately still going to take some time.

Since write_dir was added in pull #492, this should be a minor version bump (4.2.0).

The notes here don't mention how to update wfdb.readthedocs.io.  I don't remember how that's done.

tompollard:
@bemoody I bumped to v4.2.0 as suggested.

> The notes here don't mention how to update wfdb.readthedocs.io. I don't remember how that's done.

I can look into this, but would prefer to do in a new pull request.

tompollard:
@bemoody please could we merge this, so that we can release a new version?

bemoody (approved):

PR #504, opened 2024-10-09
Use uv

### LINKED ISSUE #57: Dat file compression
This may be a feature for way down the line.

Incorporate some form of file compression for dat files. One level is to use some generic compression to save space, which will increase load times. The next level would be to come up with wfdb's own compression system to load files even faster.

### LINKED ISSUE #501: Cannot initialize project with poetry
I just wanted to start developing, and after cloning the repo I did `poetry install`. After more than a minute, I received the following error(s):

<details>

```
❯ poetry install
Updating dependencies
Resolving dependencies... (66.6s)

The current project's supported Python range (>=3.7) is not compatible with some of the required packages Python requirement:
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
  - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
  - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
  - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
  - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
  - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9
  - pandas requires Python >=3.9, so it will not be satisfied for Python >=3.7,<3.9

Because no versions of pandas match >1.3.0,<1.3.1 || >1.3.1,<1.3.2 || >1.3.2,<1.3.3 || >1.3.3,<1.3.4 || >1.3.4,<1.3.5 || >1.3.5,<1.4.0 || >1.4.0,<1.4.1 || >1.4.1,<1.4.2 || >1.4.2,<1.4.3 || >1.4.3,<1.4.4 || >1.4.4,<1.5.0 || >1.5.0,<1.5.1 || >1.5.1,<1.5.2 || >1.5.2,<1.5.3 || >1.5.3,<2.0.0 || >2.0.0,<2.0.1 || >2.0.1,<2.0.2 || >2.0.2,<2.0.3 || >2.0.3,<2.1.0 || >2.1.0,<2.1.1 || >2.1.1,<2.1.2 || >2.1.2,<2.1.3 || >2.1.3,<2.1.4 || >2.1.4,<2.2.0 || >2.2.0,<2.2.1 || >2.2.1,<2.2.2 || >2.2.2,<2.2.3 || >2.2.3
 and pandas (1.3.0) requires Python >=3.7.1, pandas is forbidden.
And because pandas (1.3.1) requires Python >=3.7.1
 and pandas (1.3.2) requires Python >=3.7.1, pandas is forbidden.
And because pandas (1.3.3) requires Python >=3.7.1
 and pandas (1.3.4) requires Python >=3.7.1, pandas is forbidden.
And because pandas (1.3.5) requires Python >=3.7.1
 and pandas (1.4.0) requires Python >=3.8, pandas is forbidden.
And because pandas (1.4.1) requires Python >=3.8
 and pandas (1.4.2) requires Python >=3.8, pandas is forbidden.
And because pandas (1.4.3) requires Python >=3.8
 and pandas (1.4.4) requires Python >=3.8, pandas is forbidden.
And because pandas (1.5.0) requires Python >=3.8
 and pandas (1.5.1) requires Python >=3.8, pandas is forbidden.
And because pandas (1.5.2) requires Python >=3.8
 and pandas (1.5.3) requires Python >=3.8, pandas is forbidden.
And because pandas (2.0.0) requires Python >=3.8
 and pandas (2.0.1) requires Python >=3.8, pandas is forbidden.
And because pandas (2.0.2) requires Python >=3.8
 and pandas (2.0.3) requires Python >=3.8, pandas is forbidden.
And because pandas (2.1.0) requires Python >=3.9
 and pandas (2.1.1) requires Python >=3.9, pandas is forbidden.
And because pandas (2.1.2) requires Python >=3.9
 and pandas (2.1.3) requires Python >=3.9, pandas is forbidden.
And because pandas (2.1.4) requires Python >=3.9
 and pandas (2.2.0) requires Python >=3.9, pandas is forbidden.
And because pandas (2.2.1) requires Python >=3.9
 and pandas (2.2.2) requires Python >=3.9, pandas is forbidden.
So, because pandas (2.2.3) requires Python >=3.9
 and wfdb depends on pandas (>=1.3.0), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties
    
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.7.1"
    For pandas, a possible solution would be to set the `python` property to ">=3.7.1"
    For pandas, a possible solution would be to set the `python` property to ">=3.7.1"
    For pandas, a possible solution would be to set the `python` property to ">=3.7.1"
    For pandas, a possible solution would be to set the `python` property to ">=3.7.1"
    For pandas, a possible solution would be to set the `python` property to ">=3.7.1"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.8"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"
    For pandas, a possible solution would be to set the `python` property to ">=3.9"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```
</details>

So the version range of pandas does not match the one from this project.

However, Poetry has several issues that make it a suboptimal choice for managing a Python project (the most serious is that is completely ignores standards agreed upon by the Python community). Therefore, would you be open to using a different tool that does support the [official `pyproject.toml` specifications](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec)?

cbrnr:
Fixes #501.

To do:

- [ ] Adapt workflow to setup access to PyPI (see https://docs.astral.sh/uv/guides/publish/#publishing-a-package)
- [ ] Optimize tests (currently, I'm using `uv pip ...`, but I think this should be replaced with `uv run ...`
- [ ] Bump any package versions and/or Python?
- [ ] Use Ruff instead of Black?

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,18 +24,17 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip poetry
-          pip install ".[dev]"
+          uv pip install ".[dev]"
       - name: Install libsndfile
         if: startsWith(matrix.os, 'ubuntu')
         run: |
           sudo apt-get install -y libsndfile1
       - name: Run tests
         run: pytest
-      - name: Validate poetry file
-        run: poetry check
       - name: Check source code format
         run: black --check --diff .
 
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Run tests
         run: uv run --extra dev pytest
       - name: Check source code format
-        run: black --check --diff .
+        run: uv run --extra dev black --check --diff .
 
   test-deb10-i386:
     runs-on: ubuntu-latest
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -19,22 +19,17 @@ jobs:
         os: [windows-latest, ubuntu-latest, macos-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11"]
     steps:
-      - uses: actions/checkout@v3
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install uv
+      - uses: actions/checkout@v4
+      - name: Setup uv
         uses: astral-sh/setup-uv@v3
-      - name: Install dependencies
-        run: |
-          uv pip install --system ".[dev]"
+      - name: Install Python ${{ matrix.python-version }}
+        run: uv python install ${{ matrix.python-version }}
       - name: Install libsndfile
         if: startsWith(matrix.os, 'ubuntu')
         run: |
           sudo apt-get install -y libsndfile1
       - name: Run tests
-        run: pytest
+        run: uv run --extra dev pytest
       - name: Check source code format
         run: black --check --diff .
 
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -28,7 +28,7 @@ jobs:
         uses: astral-sh/setup-uv@v3
       - name: Install dependencies
         run: |
-          uv pip install ".[dev]"
+          uv pip install --system ".[dev]"
       - name: Install libsndfile
         if: startsWith(matrix.os, 'ubuntu')
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,3 @@ target/
 
 # pyenv
 .python-version
-
-# Poetry
-poetry.lock
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -26,7 +26,7 @@ uv add <somepackage> --optional <somegroup>
 
 ## Creating Distributions
 
-Make sure the versions in [version.py](./wfdb/version.py) and [pyproject.toml](./pyproject.toml) are updated and kept in sync.
+Make sure to update the version in [version.py](./wfdb/version.py) accordingly.
 
 It may be useful to publish to testpypi and preview the changes before publishing to PyPi. However, the project dependencies likely will not be available when trying to install from there.
 
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -16,16 +16,13 @@ black .
 
 ## Package and Dependency Management
 
-This project uses [poetry](https://python-poetry.org/docs/) for package management and distribution.
+This project uses [uv](https://docs.astral.sh/uv/) for package management and distribution.
 
-Development dependencies are specified as optional dependencies, and then added to the "dev" extra group in the [pyproject.toml](./pyproject.toml) file.
+Development dependencies are specified as optional dependencies, at least for now and until [development dependencies](https://docs.astral.sh/uv/concepts/dependencies/#development-dependencies) become more widely used.
 
 ```sh
-# Do NOT use: poetry add <somepackage> --dev
-poetry add --optional <somepackage>
+uv add <somepackage> --optional <somegroup>
 ```
-
-The `[tool.poetry.dev-dependencies]` attribute is NOT used because of a [limitation](https://github.com/python-poetry/poetry/issues/3514) that prevents these dependencies from being pip installable. Therefore, dev dependencies are not installed when purely running `poetry install`, and the `--no-dev` flag has no meaning in this project.
 
 ## Creating Distributions
 
@@ -47,10 +44,10 @@ poetry config pypi-token.test-pypi <my-testpypi-token>
 To build and upload a new distribution:
 
 ```sh
-poetry build
+uv build
 
-poetry publish -r test-pypi
-poetry publish
+uv publish --publish-url https://test.pypi.org/legacy/
+uv publish
 ```
 
 ## Creating Documentation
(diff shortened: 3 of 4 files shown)


bemoody:
Please correct the issue link in the description above - this is issue #501, not #57.

It looks like the "uv pip install" command in `run-tests.yml` gives this error:
```
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```
I'm not familiar with uv so I'm not sure exactly what this should be doing instead.

bemoody:
As for hatchling versus poetry: a big difference is that hatchling includes the test suite and example data into the source distribution (`wfdb-*-*-*.tar.gz`).  This makes the distribution 104 MB rather than 147 KB.

I'm not sure what to do.  On the one hand, we probably *ought to be* including the test suite in the source distribution.  On the other hand, the current test suite is *vastly* larger than it needs to be.

Also, since hatchling will include all files from the working directory (except those ignored by `.gitignore`?), it's important to run `uv build` from a pristine git clone or else run `git clean -d -f -x` beforehand.  I know that's something I will forget to do, so please include that in the developers' notes.

cbrnr:
Regarding hatchling, you don't have to change the behavior if you don't want to, it just has different defaults. However, you can [manually exclude](https://hatch.pypa.io/1.12/config/build/#patterns) tests and whatever else you don't want (by default, it ignores everything in `.gitignore`). Personally, I'd rather not include the tests if they significantly increase the size, which is clearly the case here.

bemoody:
Let's not include the tests (or sample data or demos) in the sdist, then.

In my experience (having worked with a lot of different build systems, just not in Python) it is better to define what files *should* go into the package rather than what files *shouldn't*.  I think it's probably best to do that by defining explicit "include" patterns in pyproject.toml.

cbrnr:
OK, so I had to bump Python to >= 3.8, because many packages are not available for 3.7 anymore. I suggest bumping to >= 3.9, because 3.8 is already EOL (https://devguide.python.org/versions/). Of course, support for 3.12 and 3.13 would be nice, but this requires NumPy >= 2, which will hopefully be supported soon.

I excluded the `tests/` and `sample-data/` directories as well as `demo-img.png` and `demo.ipynb` for the sdist, so it is now down to 155KB. If you prefer to use include patterns, please let me know what to include.

I still have to update the developers guide and will include the note on the importance of cleaning the repository before building. I also have to find out how to set up API tokens for PyPI with uv.

tompollard:
This looks good to me.

> I excluded the tests/ and sample-data/ directories as well as demo-img.png and demo.ipynb for the sdist, so it is now down to 155KB. If you prefer to use include patterns, please let me know what to include.

@bemoody my preference would be to merge this pull request, then deal with includes in a later change.

tompollard (approved):

PR #510, opened 2024-10-16
Add capability to write signal with unique `samps_per_frame` to `wfdb.io.wrsamp`

briangow:
This PR adds the capability for writing signals with unique samples per frame (`samps_per_frame`) to `wfdb.io.wrsamp`. This is typically the function that is used to write WFDB files. This was previously only possible to do by creating a Record first and using its `wrsamp` method to do the write. 

@bemoody , I do feel like checking that the frames were uniform took a bit of extra code. I'm happy to leave it like this but also open to discussing other approaches. 

I've added a couple of tests to check that this continues to work as expected.

### DIFF
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -2933,15 +2933,19 @@ def wrsamp(
             raise Exception(
                 "When using d_signal or e_d_signal, must also specify 'fmt', 'gain', and 'baseline' fields"
             )
-    if (e_p_signal is not None or e_d_signal is not None) and samps_per_frame is None:
-            raise Exception(
-                "When passing e_p_signal or e_d_signal, you also need to specify samples per frame for each channel"
-            )
+    if (
+        e_p_signal is not None or e_d_signal is not None
+    ) and samps_per_frame is None:
+        raise Exception(
+            "When passing e_p_signal or e_d_signal, you also need to specify samples per frame for each channel"
+        )
 
     # If samps_per_frame is provided, check that it aligns as expected with the channels in the signal
     if samps_per_frame:
         # Get the number of elements being passed in samps_per_frame
-        samps_per_frame_length = len(samps_per_frame) if isinstance(samps_per_frame, list) else 1
+        samps_per_frame_length = (
+            len(samps_per_frame) if isinstance(samps_per_frame, list) else 1
+        )
         # Get properties of the signal being passed
         first_valid_signal = next(
             signal for signal in signal_list if signal is not None
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -2934,15 +2934,15 @@ def wrsamp(
             )
 
     # If samps_per_frame is a list, check that it aligns as expected with the channels in the signal
-    if len(samps_per_frame) > 1:
+    if isinstance(samps_per_frame, list):
         # Get properties of the signal being passed
-        non_none_signal = next(signal for signal in signal_list if signal is not None)
-        if isinstance(non_none_signal, np.ndarray):
-            num_sig_channels = non_none_signal.shape[1]
-            channel_samples = [non_none_signal.shape[0]] * non_none_signal.shape[1]
-        elif isinstance(non_none_signal, list):
-            num_sig_channels = len(non_none_signal)
-            channel_samples = [len(channel) for channel in non_none_signal]
+        first_valid_signal = next(signal for signal in signal_list if signal is not None)
+        if isinstance(first_valid_signal, np.ndarray):
+            num_sig_channels = first_valid_signal.shape[1]
+            channel_samples = [first_valid_signal.shape[0]] * first_valid_signal.shape[1]
+        elif isinstance(first_valid_signal, list):
+            num_sig_channels = len(first_valid_signal)
+            channel_samples = [len(channel) for channel in first_valid_signal]
         else:
             raise TypeError("Unsupported signal format. Must be ndarray or list of lists.")
 
--- a/wfdb/io/record.py
+++ b/wfdb/io/record.py
@@ -2933,9 +2933,15 @@ def wrsamp(
             raise Exception(
                 "When using d_signal or e_d_signal, must also specify 'fmt', 'gain', and 'baseline' fields"
             )
+    if (e_p_signal is not None or e_d_signal is not None) and samps_per_frame is None:
+            raise Exception(
+                "When passing e_p_signal or e_d_signal, you also need to specify samples per frame for each channel"
+            )
 
-    # If samps_per_frame is a list, check that it aligns as expected with the channels in the signal
-    if isinstance(samps_per_frame, list):
+    # If samps_per_frame is provided, check that it aligns as expected with the channels in the signal
+    if samps_per_frame:
+        # Get the number of elements being passed in samps_per_frame
+        samps_per_frame_length = len(samps_per_frame) if isinstance(samps_per_frame, list) else 1
         # Get properties of the signal being passed
         first_valid_signal = next(
             signal for signal in signal_list if signal is not None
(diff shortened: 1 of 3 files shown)


wfdb/io/record.py
```diff
@@ -2911,59 +2921,91 @@ def wrsamp(
...
         if fmt is None or adc_gain is None or baseline is None:
             raise Exception(
-                "When using d_signal, must also specify 'fmt', 'gain', and 'baseline' fields."
+                "When using d_signal or e_d_signal, must also specify 'fmt', 'gain', and 'baseline' fields"
             )
-    # Depending on whether d_signal or p_signal was used, set other
-    # required features.
+
+    # If samps_per_frame is a list, check that it aligns as expected with the channels in the signal
+    if len(samps_per_frame) > 1:
```
tompollard, line 2937: If `samps_per_frame` is not a list (e.g. it's an `int`), this will error out.

>> len(4)
>> TypeError: object of type 'int' has no len()

wfdb/io/record.py
```diff
@@ -2911,59 +2921,91 @@ def wrsamp(
...
+            num_sig_channels = non_none_signal.shape[1]
+            channel_samples = [non_none_signal.shape[0]] * non_none_signal.shape[1]
+        elif isinstance(non_none_signal, list):
+            num_sig_channels = len(non_none_signal)
+            channel_samples = [len(channel) for channel in non_none_signal]
+        else:
+            raise TypeError("Unsupported signal format. Must be ndarray or list of lists.")
+
+        # Check that the number of channels matches the number of samps_per_frame entries
+        if num_sig_channels != len(samps_per_frame):
```
tompollard, line 2950: as above, `len(samps_per_frame)` will error if `samps_per_frame` is an int.

wfdb/io/record.py
```diff
@@ -2911,59 +2921,91 @@ def wrsamp(
...
-                "When using d_signal, must also specify 'fmt', 'gain', and 'baseline' fields."
+                "When using d_signal or e_d_signal, must also specify 'fmt', 'gain', and 'baseline' fields"
             )
-    # Depending on whether d_signal or p_signal was used, set other
-    # required features.
+
+    # If samps_per_frame is a list, check that it aligns as expected with the channels in the signal
+    if len(samps_per_frame) > 1:
+        # Get properties of the signal being passed
+        non_none_signal = next(signal for signal in signal_list if signal is not None)
```
tompollard, line 2939: Instead of `non_none_signal`, what about something like "first_valid_signal"?

wfdb/io/record.py
```diff
@@ -2911,59 +2921,91 @@ def wrsamp(
     if "." in record_name:
         raise Exception("Record name must not contain '.'")
     # Check input field combinations
-    if p_signal is not None and d_signal is not None:
+    signal_list = [p_signal, d_signal, e_p_signal, e_d_signal]
+    signals_set = sum(1 for var in signal_list if var is not None)
+    if signals_set != 1:
```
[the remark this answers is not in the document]
briangow: @tompollard , if all signals are passed as None a failure will occur further down the line. @bemoody , do we want to allow all signals to be None when calling `wfdb.io.wrsamp`? If we don't want to allow that then I will leave the code as I have it which requires that one of the signals is set.

wfdb/io/record.py
briangow: Fixed

wfdb/io/record.py
briangow: Fixed

wfdb/io/record.py
briangow: I switched this to `first_valid_signal` .

briangow:
@tompollard , if this failing test is of concern, could you guide me on what to do to resolve it:

```
would reformat /home/runner/work/wfdb-python/wfdb-python/wfdb/io/record.py
2 files would be reformatted, 36 files would be left unchanged.
```

wfdb/io/record.py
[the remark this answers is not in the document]
tompollard: If passing `None` for both signals (`p_signal` and `d_signal`) previously resulted in an error, then the new behavior is an improvement. Unless people were intentioanlly calling the function with both signals set to None previously, this change isn't a problem.

tompollard:
It looks like the style checks are failing (https://github.com/MIT-LCP/wfdb-python/actions/runs/11408864264/job/31747939807?pr=510). To fix, you'll need to make sure that the code conforms to black (install black, run black over the code to reformat).

bemoody:
The following raises an error:
```
>>> wfdb.wrsamp("xxx", fs=500, units=["mV"], sig_name=["I"], fmt=["16"],
                e_p_signal=[numpy.array([0.0, 1.0, 2.0])],
                adc_gain=[1.0], baseline=[0])
TypeError: 'NoneType' object is not subscriptable
```
In this case, wrsamp should raise an exception to say that if `e_p_signal` is provided, then `samps_per_frame` must be too.

bemoody:
The following raises an error:
```
>>> wfdb.wrsamp("xxx", fs=500, units=["mV"], sig_name=["I"], e_p_signal=[numpy.array([0.0, 1.0, 2.0])], fmt=["16"], samps_per_frame=[3])
TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
The problem here is that `calc_adc_params` doesn't handle the expanded case.  `calc_adc_params` should use the same method to generate `adc_gain` and `baseline` as it does in the non-expanded case.

I think it should be straightforward to restructure `calc_adc_params` to take an `expanded` argument and to use `e_p_signal` instead of `p_signal` where appropriate.

briangow:
> The following raises an error:
> 
> ```
> >>> wfdb.wrsamp("xxx", fs=500, units=["mV"], sig_name=["I"], fmt=["16"],
>                 e_p_signal=[numpy.array([0.0, 1.0, 2.0])],
>                 adc_gain=[1.0], baseline=[0])
> TypeError: 'NoneType' object is not subscriptable
> ```
> 
> In this case, wrsamp should raise an exception to say that if `e_p_signal` is provided, then `samps_per_frame` must be too.

@bemoody , thanks for catching this. I think this also applies for `e_d_signal`. I've added a check to make sure `samps_per_frame` is supplied when writing these expanded signals. 

I've also updated the check to make sure `samps_per_frame` aligns with the channels in the signal, to consider the case when `samps_per_frame` is an `int` instead of a `list`.

Finally, I think we should allow passing `samps_per_frame` with a `1` for each channel to `wrsamp` when a mixed frequency signal is read in with `smooth_frames=True`:

`rec = wfdb.io.rdrecord('/wfdb-python/sample-data/mixedsignals', smooth_frames=True)`

`print(rec.samps_per_frame)`
`[4, 4, 4, 2, 2, 1]`

`wfdb.io.wrsamp('written_mixedsignals', fs = rec.fs, units=rec.units, sig_name=rec.sig_name,
            base_date=rec.base_date, base_time=rec.base_time, comments=rec.comments, p_signal=rec.p_signal,
            samps_per_frame=[1,1,1,1,1,1], baseline=rec.baseline, adc_gain=rec.adc_gain,
            fmt=rec.fmt)`

Please let me know if you disagree, or have other thoughts with respect to this case.

briangow:
> The following raises an error:
> 
> ```
> >>> wfdb.wrsamp("xxx", fs=500, units=["mV"], sig_name=["I"], e_p_signal=[numpy.array([0.0, 1.0, 2.0])], fmt=["16"], samps_per_frame=[3])
> TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
> ```
> 
> The problem here is that `calc_adc_params` doesn't handle the expanded case. `calc_adc_params` should use the same method to generate `adc_gain` and `baseline` as it does in the non-expanded case.
> 
> I think it should be straightforward to restructure `calc_adc_params` to take an `expanded` argument and to use `e_p_signal` instead of `p_signal` where appropriate.

@bemoody , this is being addressed in a separate PR: https://github.com/MIT-LCP/wfdb-python/pull/512 .

bemoody (approved):
Looks good

PR #511, opened 2024-10-23
Bump dependencies for NumPy 2 compatibility

cbrnr:
All tests pass and `demo.ipynb` works without any errors or warnings using NumPy >= 2.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,14 +13,14 @@ env:
 
 jobs:
   test:
-    name: Python ${{ matrix.python-version }} on ${{ matrix.os }} with ${{ matrix.numpy }}
+    name: Python ${{ matrix.python-version }} on ${{ matrix.os }}
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
-        numpy: ["numpy"]
         include:
-          - python-version: "3.9"
+          - os: ubuntu-latest
+            python-version: "3.9"
             numpy: "numpy==1.26.4"
     runs-on: ${{ matrix.os }}
     steps:
@@ -35,7 +35,9 @@ jobs:
           sudo apt-get install -y libsndfile1
       - name: Install dependencies
         run: |
-          uv pip install ".[dev]" ${{ matrix.numpy }}
+          uv pip install ".[dev]" 
+      - if: ${{ matrix.numpy }}
+        run: uv pip install ${{ matrix.numpy }}
       - name: Run tests
         run: pytest
       - name: Check style
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   test:
-    name: Python ${{ matrix.python-version }} on ${{ matrix.os }}${{ matrix.numpy != 'numpy' && ' with ' + matrix.numpy || '' }}
+    name: Python ${{ matrix.python-version }} on ${{ matrix.os }} with ${{ matrix.numpy }}
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,19 +16,15 @@ jobs:
     name: Python ${{ matrix.python-version }} with ${{ matrix.numpy }}
     strategy:
       matrix:
-        #os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
-        numpy: ["numpy"]
-        include:
-          - python-version: "3.9"
+        numpy:
+          - "numpy"
+          - "numpy==1.26.4"
+        exclude:
+          - python-version: "3.13"
             numpy: "numpy==1.26.4"
-          - python-version: "3.10"
-            numpy: "numpy==1.26.4"
-          - python-version: "3.11"
-            numpy: "numpy==1.26.4"
-          - python-version: "3.12"
-            numpy: "numpy==1.26.4"
-    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


cbrnr:
OK, I'm struggling a bit with dependencies here. If you want to keep Python >= 3.9 and at the same time also support 3.13, scipy >= 1.13.1 works only for Python <= 3.12. Strangely, uv does not use scipy 1.14.1 (there is no binary wheel for scipy 1.13.1 for Python 3.13) for Python 3.13, but insists on using 1.13.1...

bemoody:
Sure, seems like the problem is that "uv run --extra dev pytest" is doing something bizarre.

I don't know exactly what that command is doing or trying to do, but it seems to be trying to install the *oldest* versions of packages that satisfy our dependencies.  That is not what we want (that's not what any normal user does) and is not likely to lead to a good solution.

cbrnr:
It might actually be something at the core of how if resolves dependencies, yes. I've opened an issue here: https://github.com/astral-sh/uv/issues/8492

YmirKhang:
Hells yeah!

cbrnr:
Note that we could also force specific NumPy versions for each Python version as shown e.g. here: https://github.com/the-siesta-group/edfio/blob/main/.github/workflows/test.yml

WDYT?

bemoody:
I appreciate your efforts to try to get to the bottom of these issues!

However, I don't think wfdb package should *require* numpy 2 at this stage, and I'm not inclined to change that just to appease uv.

Furthermore, before removing the "numpy < 2" requirement from pyproject.toml, I would like to see that the test suite is able to run on numpy 2, *with* NPY_PROMOTION_STATE set to "weak_and_warn", *and* with no warnings generated.  Currently that's not the case.

That aside, we really need to have a CI workflow that tests *whether the package can be installed using pip*.  If uv doesn't install exactly the same packages that pip would have installed, then uv is not an effective substitute.

cbrnr:
> However, I don't think wfdb package should _require_ numpy 2 at this stage, and I'm not inclined to change that just to appease uv.

Of course, that's why I suggested to pin several NumPy versions (e.g., the oldest supported NumPy version per Python version).

> Furthermore, before removing the "numpy < 2" requirement from pyproject.toml, I would like to see that the test suite is able to run on numpy 2, _with_ NPY_PROMOTION_STATE set to "weak_and_warn", _and_ with no warnings generated. Currently that's not the case.

Sure, I can do that!

> That aside, we really need to have a CI workflow that tests _whether the package can be installed using pip_. If uv doesn't install exactly the same packages that pip would have installed, then uv is not an effective substitute.

Yes, we can use the `uv pip` interface instead, which behaves identically to pip. Although Astral devs are currently working hard to implement a resolver option to `uv run/sync`, which also behaves like pip, but I don't know how long it will take them. We can immediately switch to `uv pip` though.

cbrnr:
If you are completely unhappy with uv, I can also get rid of it again and replace it with pip, just let me know.

bemoody:
Thanks!

> Of course, that's why I suggested to pin several NumPy versions (e.g., the oldest supported NumPy version per Python version).

Yeah, that seems like a reasonable workaround as long as it's not too much work.

> Yes, we can use the uv pip interface instead, which behaves identically to pip.

Sounds like that's the best thing for now.

cbrnr:
OK, I'd really appreciate if someone took a look. I've now enabled `weak_and_warn`, and the Python 3.9 job on Ubuntu uses `numpy==1.26.4` (all other jobs use `numpy>=2`). There are a couple of warnings, but also a test fails.

cbrnr:
Since uv has a new solver which now by default behaves like pip, I'll revert to using `uv sync` and `uv run` again (will take a little time though). I'd appreciate any responses to my questions, because these are not affected by the resolution strategy.

cbrnr:
Alright, this seems to be working now. @bemoody @tompollard it would be great if you could take a look and merge if you're happy with the changes. A quick follow-up release with these changes would then be super useful too! Thanks!

wfdb/processing/peaks.py
```diff
@@ -4,7 +4,7 @@
 
 
 def find_peaks(sig):
-    """
+    r"""
```
tompollard, line 7: typo?

wfdb/io/convert/edf.py
```diff
@@ -399,7 +399,7 @@ def read_edf(
         }
 
     sig_data = np.empty((sig_len, n_sig))
-    temp_sig_data = np.fromfile(edf_file, dtype=np.int16)
+    temp_sig_data = np.fromfile(edf_file, dtype=np.int16).astype("int64")
```
tompollard, line 402: edf.py was updated in https://github.com/MIT-LCP/wfdb-python/pull/527 so this change is not needed.

.github/workflows/test.yml
```diff
@@ -10,30 +8,38 @@ on:
     branches:
       - main
 
-# Set the language, install dependencies, and run the tests
+env:
+  NPY_PROMOTION_STATE: weak_and_warn
```
tompollard, line 12: When running pytest with `NPY_PROMOTION_STATE: weak_and_warn`,  I get an overflow error for Numpy < 2. Presumably some of the tests will fail here too, if Numpy < 2 is in the Numpy matrix.

tompollard (commented):
Thanks @cbrnr, just added a few quick comments.

wfdb/processing/peaks.py
cbrnr: No, this has to be a raw string because it contains the sequence `\/`, which otherwise would be interpreted as an invalid escape sequence.

wfdb/io/convert/edf.py
cbrnr: Got it!

.github/workflows/test.yml
cbrnr: The job "Test / Python 3.9 / ubuntu-latest / numpy==1.26.4" should test this.

tompollard:
Looks great, thanks again for your help @cbrnr!

tompollard (approved):

PR #512, opened 2024-10-29
Allow expanded physical signal in `calc_adc_params`

briangow:
This PR updates `calc_adc_params` to allow an expanded physical signal to be passed. Previously only a non-expanded signal was allowed. The function will now calculate the `adc_gain` and `baseline` for `p_signal` or `e_p_signal`. 

I check for the signal that is not `None` and then compute the gain and baseline. I moved the calculation itself to a new function `calc_adc_gain_baseline` so we could call it for either signal type.

### DIFF
--- a/wfdb/io/_signal.py
+++ b/wfdb/io/_signal.py
@@ -830,7 +830,9 @@ class SignalMixin(object):
             maxvals = np.nanmax(self.p_signal, axis=0)
 
             for ch in range(np.shape(self.p_signal)[1]):
-                adc_gain, baseline = self.calc_adc_gain_baseline(ch, minvals, maxvals)
+                adc_gain, baseline = self.calc_adc_gain_baseline(
+                    ch, minvals, maxvals
+                )
                 adc_gains.append(adc_gain)
                 baselines.append(baseline)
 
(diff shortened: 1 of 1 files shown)


bemoody:
Thanks!

It could be optimized somewhat: calculating min/max using a generator expression
```
minvals.append(min(x for x in ch if not math.isnan(x)))
```
will be slower than doing it with numpy.  But the important thing is that it works.

It would be good to add a test case; currently it looks like there is no test coverage for `calc_adc_params`.  The easiest way to test it is using `wfdb.wrsamp`.  So we should do that as a follow-up after merging this and #510.

briangow:
> Thanks!
> 
> It could be optimized somewhat: calculating min/max using a generator expression
> 
> minvals.append(min(x for x in ch if not math.isnan(x)))
> will be slower than doing it with numpy. But the important thing is that it works.
> 
> It would be good to add a test case; currently it looks like there is no test coverage for calc_adc_params. The easiest way to test it is using wfdb.wrsamp. So we should do that as a follow-up after merging this and https://github.com/MIT-LCP/wfdb-python/pull/510.

@bemoody , I have updated the code to use `np.nanmin` and `np.nanmax` instead of the generator function, as suggested. 

I can look into adding a test once these PRs are merged.

bemoody (approved):
This looks good!

PR #514, opened 2024-11-04
Corrected typo and extended allowed types for MultiSegmentRecord

agent3gatech:
Corrected misordered arguments in util.lines_to_file. Added 'list' as allowed types for seg_name and seg_len, for when writing MultiSegmentRecords.

### DIFF
--- a/wfdb/io/_header.py
+++ b/wfdb/io/_header.py
@@ -120,8 +120,8 @@ SEGMENT_SPECS = pd.DataFrame(
     columns=_SPECIFICATION_COLUMNS,
     dtype="object",
     data=[
-        [(str), "", None, True, None, None],  # seg_name
-        [int_types, " ", "seg_name", True, None, None],  # seg_len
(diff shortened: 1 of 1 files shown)


bemoody:
Thanks!

In case it's not clear, here's an example which doesn't work prior to this change:

```
>>> import wfdb
>>> r = wfdb.rdheader('3000003', pn_dir='mimic3wdb/1.0/30/3000003')
>>> r.wrheader()
```

Probably nobody tested this code path, and the logic is hard to follow.  This change should fix the immediate problem.

But I think what we really want is, rather than changing the types in `SEGMENT_SPECS`, we want to add `"seg_name"` and `"seg_len"` to `LIST_FIELDS` (wfdb/io/record.py, line 1675.)

briangow:
Thanks @agent3gatech !

briangow (approved):
This will solve the problem! I'll open an issue to capture the alternate suggestion by @bemoody .

PR #519, opened 2024-11-27
Fix selection of channels when converting to EDF

### LINKED ISSUE #518: Unable to choose channels when converting to EDF.
the `wfdb-to_edf()` function takes an optional `channels` argument, which should allow you to specify which channels you want to be read. This doesn't work currently as it is not passed to `rdrecord`, so irrespective of what you pass into the function the created EDF file has all of the channels included.

SamJelfs:
Fixes #518


tompollard:
Thanks @SamJelfs!

tompollard (approved):

Issue #525, opened 2025-01-16
The ricker wavelet is removed from scipy.signal from version 1.15


wenh06:
The `ricker` wavelet was removed from `scipy.signal` in version 1.15, released on 2025.1.3. This affects the [QRS detection functions](https://github.com/MIT-LCP/wfdb-python/blob/main/wfdb/processing/qrs.py).

wenh06:
@briangow @bemoody

tompollard:
Fixed in https://github.com/MIT-LCP/wfdb-python/pull/526, thanks!

PR #526, opened 2025-01-16
copy ricker wavelet from scipy which removed it from version 1.15.0

wenh06:
This PR fixes issue #525

### DIFF
--- a/wfdb/processing/qrs.py
+++ b/wfdb/processing/qrs.py
@@ -215,7 +215,7 @@ class XQRS(object):
         N/A
 
         """
-        wavelet_filter = signal.ricker(self.qrs_width, 4)
+        wavelet_filter = ricker(self.qrs_width, 4)
 
         self.sig_i = (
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks @wenh06! If copying, we need to make sure that we follow the license terms, which I think are here: https://github.com/scipy/scipy/blob/main/LICENSE.txt

Alternatively, we could add a dependency on the recommended package: https://pywavelets.readthedocs.io/en/latest/

wenh06:
The [ricker wavelet]() is defined as

$$
\psi (t)={\frac {2}{{\sqrt {3\sigma }}\pi ^{1/4}}}\left(1-\left({\frac {t}{\sigma }}\right)^{2}\right)e^{-{\frac {1}{2} \left({\frac {t}{\sigma }}\right)^{2}}}
$$

But `pywt` seems to fix $\sigma = 1$, I don't know how to change this parameter with `pywt`.

bemoody (approved):

tompollard:
Looks good to me, thanks again @wenh06

PR #544, opened 2025-06-23
Update `MultiHeaderMixin` header formatting to match `HeaderMixin` formatting

briangow:
In 63c987f33098aa0ccded214115d7922a7d0b9325 updates to `HeaderMixin` were made to properly format the header date and time fields. This fix did not make it into `MultiHeaderMixin` (for multi-segment records). Therefore, writing a date using the `wrheader` method under `MultiHeaderMixin` currently results in writing the `base_date` as YYYY-MM-DD which isn't compatible with the DD/MM/YYYY format requirement. 

This PR updated `MultiHeaderMixin` by using the code from `HeaderMixin` to properly set formats.

### DIFF
--- a/wfdb/io/_header.py
+++ b/wfdb/io/_header.py
@@ -1,4 +1,5 @@
 import datetime
+from dateutil import parser
 from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


bemoody:
Not tested but it makes sense to me. Thanks!

The logic for reformatting date strings looks fragile and I wonder if there's a reason it's not just using strftime("%d/%m/%Y"). But that's another issue.

briangow:
@bemoody , I agree that the old approach of formatting the `base_date` seemed fragile. I've updated this to use `.strftime`.

PR #548, opened 2025-10-13
Fix `OverflowError: Python int too large to convert to C long` raised during annotation label map creation with custom labels.

briangow:
The patch makes the label‑store dtype explicit (int64) and replaces row‑wise .loc assignment with a keyed, vectorized update.


bemoody:
Pandas is Greek to me. What is the issue this is trying to solve?

What integer value is too large to convert to long? Annotation *type codes* can't be too large; those must be between 1 and 49.

PR #550, opened 2025-11-25
Add support for european dates in header files.

Ivorforce:
Modeled after `wfdb_strptime` (directly above).

I'm not sure if this is violating the spec, but it should be fairly unambiguous with this date format (moreso than the `d/m/y` format, which might be interpreted as `m/d/y`).


Ivorforce:
Just noticing that this change wouldn't work just yet because of the regex above. I'll wait for feedback before fixing that though.

Issue #554, opened 2026-01-07
wfdb.io.MultiRecord.wrsamp missing option for expanded=True


agent3gatech:
wfdb.io.MultiRecord.wrsamp (which just calls 'self.wrheader' and then 'for seg in self.segments: seg.wrsamp') seems to be missing the ability to pass 'expanded=True', as one can with wfdb.io.Record.wrsamp. 

This causes checksums to fail as the checks look for things related to 'd_signal' rather than 'e_d_signal', for example.

PR #555, opened 2026-01-08
Allow expanded parameter in `MultiRecord.wrsamp`

briangow:
As raised in #554, `wfdb.io.Multirecord.wrsamp` does not provide the option to pass `expanded=True`. This should be provided so that multifrequency signals can be written within the `Multirecord` class. This PR updates the `wrsamp` call to allow for `expanded` to be passed.


briangow:
Thanks @bemoody! I agree with your comment about additional validation. I've opened issue #557 to capture that. We can use separate PRs to implement those validations.

